PR #10, opened 2016-04-08
feat(tests): adding a test suite

bcoe:
* adds a test-suite, inspired by the integration tests in `standard-changelog`.
* adds support for `Node 0.10`.
* adds a header, _marketing for the win_

### DIFF
--- a/index.js
+++ b/index.js
@@ -105,7 +105,12 @@ function outputChangelog (argv, cb) {
 }
 
 function commit (argv, newVersion, cb) {
-  checkpoint('commiting %s and %s', ['package.json', 'argv.infile'])
+  var msg = 'committing %s', args = [argv.infile]
+  if (!argv.firstRelease) {
+    msg += ' and %s'
+    args.unshift('package.json')
+  }
+  checkpoint(msg, args)
   exec('git add package.json ' + argv.infile + ';git commit package.json ' + argv.infile + ' -m "' + formatCommitMessage(argv.message, newVersion) + '"', function (err, stdout, stderr) {
     var errMessage = null
     if (err) errMessage = err.message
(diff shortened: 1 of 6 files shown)


nexdrew:
How about we go ahead and add a `"files"` stanza to package.json?

```json
"files": [
  "index.js"
]
```

Other than that, I think this looks good. :+1:

nexdrew:
Whoops, actually, this doesn't look right:

```
✔ commiting package.json and argv.infile
```

1. Prints `argv.infile` instead of `CHANGELOG.md`
2. I think "commiting" [should be spelled with two t's](https://www.google.com/search?q=define+commiting)

index.js
```diff
@@ -75,43 +73,39 @@ conventionalRecommendedBump({
...
+  changelogStream.on('end', function () {
+    checkpoint('outputting changes to %s', [argv.infile])
+    fs.writeFileSync(argv.infile, header + '\n' + content + oldContent, 'utf-8')
+    return cb()
+  })
 }
 
 function commit (argv, newVersion, cb) {
-  console.log(chalk.bold('3.') + ' commit ' + chalk.bold('package.json') + ' and ' + chalk.bold(argv.infile))
+  checkpoint('commiting %s and %s', ['package.json', 'argv.infile'])
```
nexdrew: Here's the culprit, change `'argv.infile'` to `argv.infile`.

Also, note that when using `--first-release` this checkpoint says it is committing package.json, but package.json is not actually modified (no version bump). Might want to make the checkpoint message and args conditional:

```js
var msg = 'committing %s', args = [argv.infile]
if (!argv.firstRelease) {
  msg += ' and %s'
  args.push('package.json')
}
checkpoint(msg, args)
```

or the less verbose:

```js
checkpoint(argv.firstRelease ? 'commiting %s' : 'commiting %s and %s', argv.firstRelease ? [argv.infile] : ['package.json', argv.infile])
```

Issue #15, opened 2016-04-10
Support for tag signing


Conaclos:
Hi! Very nice project!

I'm currently signing my tags (`git tag -s`).
It could be nice to add an option to enforce tag signing.
Similar to the npm's option `sign-git-tag`.

bcoe:
@Conaclos this sounds like a great addition \o/ would you like to take a stab at adding it?

The unit tests are pretty easy to write, they're simply git integration tests.

Conaclos:
Sorry, what's a **stab**?

After a bit of investigation, I found the addition of a new feature in the coming release of Git.

I wrote more details about it in #18.
Although the PR is ridiculously little, it makes **standard-version** ready for Git 2.9 ;)

bcoe:
@Conaclos `stab` is a saying meaning basically "attempt to implement". 

I think it would be nice to have this added as a command line option, I'll comment on the pull.

nexdrew:
@Conaclos Give `standard-version@2.2.1` a try and let us know if you have any trouble with the new `--sign` option. And thanks for your contributions!

PR #18, opened 2016-04-11
Tag signing

Conaclos:
Hi!

This PR is related to #15

`git tag -m` already creates an annotated tags, without explicitly given an
`-a` (unsigned tag) or `-s` (signed tag) option (See [git-tag man page][git-tag-man]).

From [Git 2.9][git2-9] the configuration variable `tag.forceSignAnnotated` enables to turn on
tag signing by default, in the case where `-a` is omitted.

Note: Git 2.9 is a coming release.

[git-tag-man]: https://www.kernel.org/pub/software/scm/git/docs/git-tag.html
[git2-9]: https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.9.0.txt

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,6 +24,13 @@ var argv = require('yargs')
     default: false,
     global: true
   })
+  .option('sign-tag', {
+    alias: 's',
+    describe: 'Should the git tag be signed?',
+    type: 'boolean',
+    default: false,
+    global: true
+  })
   .help()
   .alias('help', 'h')
   .example('$0', 'Update changelog and tag release')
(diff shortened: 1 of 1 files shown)


bcoe:
@Conaclos in an effort to keep the existing approach the same, but to introduce the option of signing, I suggest that we add a new command line option. perhaps `--sign-tag`:

```js
.option('sign-tag', {
  alias: 's',
  describe: 'should the git tag be signed (-s).'
  default: false
}}
```
If left false, we would append `-a` rather than `-s` to the `git tag` command.

Conaclos:
Hi!

I've just amended my commit.

I manually tested it.
However I don't see how automate the test without pollute the environment where the test is executed.

bcoe:
@Conaclos I'm happy to toss a test in, I'll submit against your branch.

bcoe:
@Conaclos haven't forgotten about this, sorry that it's taking me a while to hammer out a test.

Conaclos:
@bcoe It's ok ;)

nexdrew:
Perhaps we should sign the commit as well, via the `-S` flag? If we sign both the commit and the tag, then perhaps the CLI option should be just `--sign` (instead of `--sign-tag`) with a modified description.

bcoe:
@Conaclos are you okay with opting for #29, gets us the tests we were looking for.

Conaclos:
@bcoe @nexdrew 
I'm ok for also sign the release commit.
It is already in my workflow since, by default, I sign every commit ;)

However I know some person, e.g. [Linus Torvalds](http://git.661346.n2.nabble.com/GPG-signing-for-git-commit-td2582986.html), prefers only to sign the tags.

PR #25, opened 2016-04-26
chore(package): update shelljs to version 0.7.0

nexdrew:
Same as #23 but with compatibility fix for test.

I attempted to update the upstream branch but I do not have push access to the repo.


bcoe:
@nexdrew good find, seems like `shelljs` could benefit from `standard-version`.

PR #33, opened 2016-05-02
test: mock git for some tests to improve code coverage

nexdrew:
Mocks git (for a subset of tests) in order to better test:

- `--sign` attempts to sign the commit **and** tag
- exits with error if git commit fails
- exits with error if git tag fails

This improves code/line coverage from 91.57 to 95.18 percent.

Still need a way to induce/mock error scenarios for `conventionalRecommendedBump` and `conventionalChangelog`.

### DIFF
--- a/test.js
+++ b/test.js
@@ -6,11 +6,25 @@ var shell = require('shelljs')
 var fs = require('fs')
 var path = require('path')
 var cliPath = path.resolve(__dirname, './index.js')
+var PATH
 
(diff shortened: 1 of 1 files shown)


stevemao:
Wow this is interesting :) Can definitely be refactored into a module :)

nexdrew:
@stevemao Hmm, not a bad idea, thanks!

PR #36, opened 2016-05-07
refactor(mockGit): modulization

stevemao:
https://github.com/stevemao/mock-git

The standalone module has a few improvements:
1. It is async
2. It uses temp folder for path which is better than `cwd`
3. It also handles some potential edge cases

### DIFF
--- a/test.js
+++ b/test.js
@@ -5,8 +5,8 @@
 var shell = require('shelljs')
 var fs = require('fs')
 var path = require('path')
+var mockGit = require('mock-git')
 var cliPath = path.resolve(__dirname, './index.js')
-var PATH
 
 require('chai').should()
 
@@ -20,13 +20,6 @@ function writePackageJson (version) {
   }), 'utf-8')
 }
 
-function mockGit (logic) {
-  fs.writeFileSync('git', '#!/usr/bin/env node\n' + logic, 'utf8')
-  shell.chmod('+x', 'git')
-  PATH = shell.env.PATH
-  shell.env.PATH = shell.pwd() + ':' + PATH
-}
-
 describe('cli', function () {
   beforeEach(function () {
     shell.rm('-rf', 'tmp')
@@ -38,10 +31,6 @@ describe('cli', function () {
   })
 
   afterEach(function () {
-    if (PATH) {
-      shell.env['PATH'] = PATH
-      PATH = undefined
-    }
     shell.cd('../')
     shell.rm('-rf', 'tmp')
   })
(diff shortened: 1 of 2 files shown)


bcoe:
@stevemao @nexdrew love it.

test.js
```diff
@@ -94,36 +83,48 @@ describe('cli', function () {
   describe('with mocked git', function () {
     it('--sign signs the commit and tag', function () {
       // mock git with file that writes args to gitcapture.log
-      mockGit('require("fs").appendFileSync("gitcapture.log", JSON.stringify(process.argv.splice(2)) + "\\n", "utf8")')
-      writePackageJson('1.0.0')
+      mockGit('require("fs").appendFileSync("gitcapture.log", JSON.stringify(process.argv.splice(2)) + "\\n")')
```
bcoe: @nexdrew @stevemao clever :) a couple thoughts about `mockGit`:

* I wonder if the mocking library could be made more generic, as a tool for mocking an arbitrary bin in path (then you could instantiate one for git).
* If we opt for the git-specific approach, I wonder if we could provide a few common "logic scripts" that would fit common use-cases that folks would run into while testing git.

¯\_(ツ)_/¯

bcoe:
@nexdrew I'm cool with landing this, if you sign off.

test.js
stevemao: > I wonder if the mocking library could be made more generic, as a tool for mocking an arbitrary bin in path (then you could instantiate one for git).

I have created [mock-bin](https://github.com/stevemao/mock-bin) and mock-git is depending on it.

> I wonder if we could provide a few common "logic scripts"

There is an optional arg, [command](https://github.com/stevemao/mock-git#mockgitjs-command), you could specify. The edge case I mentioned is when your command is `commit`, it won't affect other commands such as `push`. Although it doesn't matter at the moment here.

test.js
```diff
@@ -94,36 +83,48 @@ describe('cli', function () {
...
-      captured[captured.length - 2].should.deep.equal(['tag', '-s', 'v1.0.1', '-m', 'chore(release): 1.0.1'])
+          unmock()
+        })
     })
 
     it('exits with error code if git commit fails', function () {
       // mock git by throwing on attempt to commit
-      mockGit('if (process.argv[2] === "commit") { console.error("commit yourself"); process.exit(128); }')
-      writePackageJson('1.0.0')
+      mockGit('console.error("commit yourself"); process.exit(128);', 'commit')
```
stevemao: This is an example of using `command`. If we did use `git push` here, it shouldn't be mocked @bcoe

nexdrew:
@bcoe @stevemao Cool, I like this. 👍 

I do think it'd be neat to build some conveniences around `mock-bin`, like the ability to capture args and return/fetch them from the API, but I guess that could always be another module/package. ¯\_(ツ)_/¯

Otherwise, nice work Steve!

stevemao:
Cool, thanks guys :)

PR #38, opened 2016-05-21
feat: append eol to package.json

### LINKED ISSUE #37: Add EOL when overwriting package.json
Otherwise changes to package.json b/w releases can end up flip-flopping the last line since editors like Atom will automatically add it but `standard-version` strips it off - just becomes annoying, unnecessary changes in history and in PRs.

I think we just need to change this line:

```js
fs.writeFileSync(pkgPath, JSON.stringify(pkg, null, 2), 'utf-8')
```

to something like this:

```js
fs.writeFileSync(pkgPath, JSON.stringify(pkg, null, 2) + require('os').EOL, 'utf-8')
```

nexdrew:
Fixes #37.

I'm thinking we should probably replace all `\n` characters with `os.EOL` to be more Windows friendly. I separated the commits so I can easily revert the 2nd one if you disagree.

### DIFF
--- a/index.js
+++ b/index.js
@@ -62,7 +62,7 @@ conventionalRecommendedBump({
     newVersion = semver.inc(pkg.version, release.releaseAs)
     checkpoint('bumping version in package.json from %s to %s', [pkg.version, newVersion])
     pkg.version = newVersion
-    fs.writeFileSync(pkgPath, JSON.stringify(pkg, null, 2) + eol, 'utf-8')
+    fs.writeFileSync(pkgPath, (JSON.stringify(pkg, null, 2) + '\n').replace(/\n/, eol), 'utf-8')
   } else {
     console.log(chalk.red(figures.cross) + ' skip version bump on first release')
   }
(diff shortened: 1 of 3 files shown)


bcoe:
👍 no objection from me :shipit:

bcoe:
@nexdrew made you an owner of the npm package FYI.

stevemao:
Would this cause inconsistent eol? Or we don't want to enforce it?

nexdrew:
@stevemao Not sure I follow you. What do you mean by "inconsistent eol"? Do you mean changing it b/w Windows and *nix runs?

Do you think we should try to detect which eol the files have (like Atom) before rewriting/appending to them? That would probably be better.

stevemao:
If i ran the command on windows and later on i run the same command on
linux, there would be a change because of different eol. Not sure if this
is expected?

nexdrew:
@stevemao Yeah, that's a good point. I'll change it to detect existing eol instead of always using the platform-specific one.

nexdrew:
I seem to have opened a small can of worms here. It works as expected for package.json (though we might want to consider also using [`detect-indent`](https://github.com/sindresorhus/detect-indent) to preserve indentation too) but I'm not too sure about CHANGELOG.md since `conventional-changelog` always uses `\n` (I think), which seems less important given that the content there should _always_ be generated.

If you think this is too much, I'd be ok simplifying this to just append `\n` to package.json and splitting the EOL preservation into another PR. Definitely a little scope creep here.

stevemao:
Well `conventional-changelog` uses [templates](https://github.com/conventional-changelog/conventional-changelog-writer/blob/master/templates/commit.hbs) so yeah it should be always `\n`. I never came across `require('os').EOL` before and I'm not sure about when to use it, when not to.

nexdrew:
Closing in favor of #42 (fix without scope creep) and #43 (new issue to discuss problems encountered here).

PR #53, opened 2016-06-08
test: explicitly run cli with node to fix tests on windows

### LINKED ISSUE #50: Tests fail on windows with JScript compilation errors
`npm t` never even gets to any work on Windows due to a `Microsoft JScript compilation error` being thrown during each test case.

Tapppi:
Explicitly run the cli with 'node <index.js path>' to fix running tests on windows.
There seems to be a bug overwriting project package.json and CHANGELOG.md
on windows after failed tests.

Fixes #50

### DIFF
--- a/test.js
+++ b/test.js
@@ -14,6 +14,10 @@ function commit (msg) {
   shell.exec('git commit --allow-empty -m"' + msg + '"')
 }
 
+function execCli (argString) {
+  return shell.exec('node ' + cliPath + (argString != null ? ' ' + argString : ''))
+}
+
 function writePackageJson (version) {
   fs.writeFileSync('package.json', JSON.stringify({
     version: version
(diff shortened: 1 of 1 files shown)


test.js
```diff
@@ -14,6 +14,10 @@ function commit (msg) {
   shell.exec('git commit --allow-empty -m"' + msg + '"')
 }
 
+function execCli (argString) {
+  return shell.exec('node ' + cliPath + ' ' + argString)
```
nexdrew: Should change this to:

```js
return shell.exec('node ' + cliPath + (argString ? ' ' + argString : ''))
```

To avoid command strings like `node /Users/nexdrew/git/qi/standard-version/index.js undefined`

Tapppi: Great catch! To allow falsy values without allowing undefined or null I changed it to:
```js
return shell.exec('node ' + cliPath + (argString != null ? ' ' + argString : ''))
```

stevemao: That's actually not necessary :)

Tapppi:
New `git add` test added in #55 will break this again, so probably should off on and rebase after that's merged?

test.js
nexdrew: As Steve said, I don't think we need to worry about allowing falsy values here, but this should be fine.

nexdrew:
@Tapppi Rebasing this once #55 is merged makes sense to me - good call.

Tapppi:
Rebased after #55 was merged, new test updated to use execCli as well. Should be fine for a merge.

nexdrew:
Awesome, thanks!

Issue #57, opened 2016-06-08
Include AppVeyor CI config for windows test


Tapppi:
As was pointed out by @bcoe in #55 it would be great to have AppVeyor run windows tests. Merging TravisCI and AppVeyor coverage is not so simple though (see lemurheavy/coveralls-public/issues/613).
Suggested base config was https://github.com/yargs/yargs-parser/blob/master/appveyor.yml. 

Some modifications:  
* Added core.autocrlf true to enforce CRLF line endings for the test.
* Added `nodejs_version: '0.10'` as it is present in travis.yml as well.
* Switched v5 to v6 as its the latest, is there some real need to include v5?
* Any need for `shallow_clone: true` and `clone_depth: 1`? Removed them for now.

So that leaves us with:
```
# Fix line endings on Windows
init:
  - git config --global core.autocrlf true
environment:
  matrix:
    - nodejs_version: '6'
    - nodejs_version: '4'
    - nodejs_version: '0.12'
    - nodejs_version: '0.10'
install:
  - ps: Install-Product node $env:nodejs_version
  - set CI=true
  - npm -g install npm@latest
  - set PATH=%APPDATA%\npm;%PATH%
  - npm install
matrix:
  fast_finish: true
build: off
version: '{build}'
test_script:
  - node --version
  - npm --version
  - npm test
```

On a related note, the travis.yml should probably include node version 4 as well as the latest (v6)?

PR #60, opened 2016-06-21
chore: add appveyor ci configuration

Tapppi:
I added ci.appveyor.com to this project. Not much changed since #57, removed the `core.autocrlf = true` config. Is there any need for it really?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,21 @@
+environment:
+  matrix:
+    - nodejs_version: '6'
+    - nodejs_version: '5'
+    - nodejs_version: '4'
+    - nodejs_version: '0.12'
+    - nodejs_version: '0.10'
+install:
+  - ps: Install-Product node $env:nodejs_version
+  - set CI=true
+  - npm -g install npm@latest
+  - set PATH=%APPDATA%\npm;%PATH%
+  - npm install
+matrix:
+  fast_finish: true
+build: off
+version: '{build}'
+test_script:
+  - node --version
+  - npm --version
+  - npm test


bcoe:
@Tapppi landed this so that we'd kick off an AppVeyor build, looks like we might need to do some work:

https://ci.appveyor.com/project/bcoe/standard-version/build/job/dbrpfj3f570n8h1c

Perhaps `git` isn't available?

Tapppi:
Yeah I just noticed that, I'm debugging on my fork, will report shortly.

Tapppi:
Also, the config is missing the final newline :cry: Hurts my soul that I made that mistake.

Tapppi:
So the build finished, with all the logging. Theres a bunch of git warnings about author info not being set, but I guess that doesn't matter much, except for the fact that it's annoying when debugging. The actual error seems to be `Error in options.transform: Invalid date`, happening after `√ created CHANGELOG.md`. I have absolutely no idea where this might come from. Maybe this is related to `node v6`? I'm running v4.

https://ci.appveyor.com/project/Tapppi/standard-version/build/10/job/ksaf6xchqvhavmwn

Tapppi:
I have a nagging feeling this could be related to the git log --format command used by gitRawCommit including % signs in the format argument and this % sign *thing* in appveyor https://github.com/appveyor/ci/issues/246 somehow.. :disappointed: Still not sure though. I'll look into it more when I can.

bcoe:
@Tapppi tested on Node V6 on OSX, works BTW -- wish AppVeyor had better output. We should look at the npm/npm appveyor config, I think there are some tricks to get better output.

stevemao:
@bcoe Can you add a badge to README? appveyor config is per user instead of per org... I can point the CI under my name to this repo... Which is weird ..

bcoe:
@stevemao 👍 will add as soon as we hammer out this weird issue... yeah, appveyor doesn't do orgs too well.

Tapppi:
So I put a debug build together. It basically has a node_modules.tar.gz that has all deps with debug logs inserted. As you can see, the git log output has a value of `truen` for `committerDate` (git log format specifier `%ci`) for some reason (the actual line is `344`):  
https://ci.appveyor.com/project/Tapppi/standard-version/build/19/job/v0a2h31kk5b97e73#L337

I have no idea how that can be :disappointed:

Tapppi:
Here is another test from earlier, where I tested `git log --format` (executed in `powershell`) before running the tests with the format string that's used in tests. The output is just fine..  
https://ci.appveyor.com/project/Tapppi/standard-version/build/15/job/hc2a86i2va2mjsxa#L881

bcoe:
@Tapppi I have quite a few cycles to put into OSS this weekend, so I can dive in and help debug too; if this is still stumping us.

Tapppi:
I forgot to check the last test i put together and, lo and behold, it got the most interesting results so far. For the heck of it I put another debug together, where the `git log` `--format` string is always just logging all the dates (`%ci %cI %ai %aI`). 

The committer dates (`%ci %cI`), which are used, are always `truen` even though the author dates are correct:

https://ci.appveyor.com/project/Tapppi/standard-version/build/20/job/piotnr784candb99#L592

As far as I can see, this could be either:
* `git`'s intended behavior manifesting due to some strange way the tests are run
* a bug in the specific version of `git` in Appveyor 

I can't find any documentation pointing out that `truen` could be a value for `%ci` though..

PR #62, opened 2016-06-23
fix: try installing newer version of git

bcoe:


Tapppi:
Found it, this is definitely caused by the CI environment variable. The `%ci%` in the format is interpreted as an env var. Whether we set it or not doesn't matter since Appveyor sets it automatically. See this build where I `set CI=2016-06-29` and the resulting log from format `%ci%n...%cI%n` has that as output and is missing the `%n` newline, since the percent sign is interpreted as part of the `%ci%` var.

https://ci.appveyor.com/project/Tapppi/standard-version/build/22/job/at1l9hg6gpvpcaqr#L361

Tapppi:
Doesn't seem like there is a way to clear the env var.. Maybe fix this by including a space after the `%ci` var and before the newline in the format? That would need to go in `conventional-changelog-core` defaults or the calling options in this package though, whichever. I can create a PR after we decide which it should go on.

Fixed with a space after the `%ci` tag:
https://ci.appveyor.com/project/Tapppi/standard-version/build/25/job/xb9r1g3tk312tfej

Many ways tried  to set the CI env var to nothing, to no avail:
https://ci.appveyor.com/project/Tapppi/standard-version/build/24/job/ua3jmju88ykyeoqp#L293

nexdrew:
Just to clarify (mainly for my own benefit):

1. The [default `git log` format](https://github.com/conventional-changelog/conventional-changelog-core/blob/v1.5.0/lib/merge-config.js#L195) used by `conventional-changelog-core` (passed to `git-raw-commits`) causes problems on AppVeyor because a `%ci%` in the format is expanded to the value of the `CI` env var in Windows (?)

    This breaks on AppVeyor but shouldn't break for normal Windows users that don't have a `CI` env var defined, correct?

    Perhaps we should add AppVeyor/Windows testing to `conventional-changelog-core`?

2. `mock-git` (or maybe `mock-bin`) does not currently work on Windows, as [represented in Tapppi's build](https://ci.appveyor.com/project/Tapppi/standard-version/build/25/job/xb9r1g3tk312tfej#L946).

    This would affect the tests on Windows but should not keep `standard-version` from working for normal Windows users.

    Perhaps we should add AppVeyor/Windows testing to `mock-bin` and `mock-git`?

So, besides the test suite currently failing on Windows (due to no. 2) and in AppVeyor (due to no. 1 and no. 2), do we know what actually blocks folks from using `standard-version` on Windows?

PR #64, opened 2016-06-23
fix(index.js): use blue figures.info for last checkpoint

nexdrew:
So it doesn't look like something that was already done.

Before:
screen shot 2016-06-23 at 12 34 56 pm

After:
screen shot 2016-06-23 at 12 35 40 pm

Just a minor change.


Tapppi:
LGTM!

nexdrew:
@Tapppi Thanks for the review!

PR #75, opened 2016-07-25
feat: preset support

azachar:
Hi there, 

I implemented support for the preset parameters. Now you can specify the param --preset that is passed to  ``conventional-changelog``

For some reason, I've got stuck. The ``require`` cannot find ``conventional-changelog-<<name>>`` in ``conventional-changelog`` module.

What do you suggest to fix it?

For instance, do I need to install the module ``conventional-changelog-atom`` as a global dependency?

Thanks for your reply.

Cheers,
Andrej

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,17 @@ _but wait, there's more!_
 Github usernames (`@bcoe`) and issue references (#133) will be swapped out for the
 appropriate URLs in your CHANGELOG.
 
+### Preset
+
+To generate your changelog with a different message format preset, simply use:
+
+```sh
+standard-version --preset atom
+```
+
+You can use as the same preset names as in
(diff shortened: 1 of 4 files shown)


stevemao:
I don't think we need this since the point of this module is to be opinionated.

Tapppi:
@stevemao is there a better alternative to point to? If not, this becomes a bigger question.

stevemao:
We don't really recommend people to use other conventions. But modules like https://github.com/conventional-changelog/conventional-changelog-cli supports it.

Tapppi:
Yeah I just realised that the cli would probably be the best alternative.

bcoe:
@azachar, as @stevemao points out, we're trying to keep this module grounded in the `angular-commit-format`; but, give `conventional-changelog-cli` a spin 👍

Issue #73, opened 2016-07-15
Use listr?


stevemao:
[listr](https://github.com/SamVerschueren/listr) looks like a good package to use for our task list. We should look into it :)

bcoe:
closing this, as we opted to use our own checkpoint implementation.

PR #81, opened 2016-08-03
refactor to use listr for tasks

dignifiedquire:
Based on #73 this refactors the code to use [Listr](https://github.com/SamVerschueren/listr) for task display. I tried to keep the output as close as possible to what is there currently. 

The git mock tests are failing as I think using [execa](https://github.com/sindresorhus/execa) it does not detect the mocked `git` but rather the original.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,6 @@
 #!/usr/bin/env node
-var conventionalRecommendedBump = require('conventional-recommended-bump')
-var conventionalChangelog = require('conventional-changelog')
-var path = require('path')
+
+var tasks = require('./tasks')
 var argv = require('yargs')
   .usage('Usage: $0 [options]')
   .option('infile', {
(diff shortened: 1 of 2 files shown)


stevemao:
I was playing with listr too and it's really nice. There's a few features missing EG we can't a customize figure for certain task (we want `figures.info` instead of `figures.tick` for the last task). I'll look into this soon :)

nexdrew:
While I like this idea, we will need to drop support for Node 0.10 and 0.12 to use `listr`. I'm not really opposed to this, but it needs to be a conscious decision.

@bcoe @Tapppi Thoughts?

Tapppi:
I'm +1 for using `listr` and dropping 0.10 and 0.12.

While I believe the latest stats I saw put 0.10 and 0.12 at a high share of node versions used in production, I think dropping support in `standard-version` is alright as this is a tool for publishing modules and not something running in production. Folks developing for versions <4.0 could well use >4.0 (probably with something like `nvm`) when bumping versions and publishing, and it shouldn't really affect them THAT much.

This would also cut down CI run times and simplify supporting this package. So unless someone can give valid points for keeping the support, I don't see a problem that big.

JaKXz:
I think dropping node 0.10 is perfectly valid since it will be completely unsupported soon, and 0.12 will follow soon enough.

bcoe:
@dignifiedquire I'm comfortable with us dropping `0.10` and `0.12` support, if you want to dust this work off and rebase with some of the refactoring @Tapppi has been doing (specifically #114).

dignifiedquire:
@bcoe sure I will take a look tomorrow :)

bcoe:
@dignifiedquire please feel free to re-open this when you have the time to pick it up ... I understand how free time to work on OSS can be in short supply.

I'm doing a bit of spring cleaning so that we don't get buried in open pull requests on this repo.

PR #83, opened 2016-08-08
Feature/add preid param for semver

stvnwrgs:
New argument `-p` allows you to create prereleases. You can also define a prerelease identifier. via `--pre-id="alpha"` (beta is default).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 .nyc_output
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 node_modules
 .nyc_output
-.idea
--- a/README.md
+++ b/README.md
@@ -105,6 +105,12 @@ npm run release -- --no-verify
 standard-version --no-verify
 ```
 
+### Create a Prerelease
+```sh
+# chanel: alpha, beta, etc
+standard-version -p --pre-id="<channel>"
+```
+
 ### CLI Help
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -107,8 +107,8 @@ standard-version --no-verify
 
 ### Create a Prerelease
 ```sh
-# chanel: alpha, beta, etc
-standard-version -p --pre-id="<channel>"
+# identifier: alpha, beta, etc
+standard-version -p --pre-id="<identifier>"
 ```
 
 ### CLI Help
--- a/README.md
+++ b/README.md
@@ -108,7 +108,15 @@ standard-version --no-verify
 ### Create a Prerelease
 ```sh
 # identifier: alpha, beta, etc
-standard-version -p --pre-id="<identifier>"
+standard-version -p --tag="<identifier>"
+```
+
+### Dry run
+If you only want to get the actual version or the next version that would be built:
+```sh
+standard-version -a
+standard-version -x
+standard-version -p --tag="<identifier>" -p
 ```
 
 ### CLI Help
--- a/index.js
+++ b/index.js
@@ -91,12 +91,11 @@ conventionalRecommendedBump({
     return
   }
   if (argv.actual) {
-    console.log(pkg.version);
+    console.log(pkg.version)
     return
   }
   if (argv.next) {
-
-    console.log(getNextVersion(release));
+    console.log(getNextVersion(release))
     return
   }
   var newVersion = pkg.version
@@ -122,7 +121,7 @@ conventionalRecommendedBump({
   }
 })
 
-function outputChangelog(argv, cb) {
+function outputChangelog (argv, cb) {
   createIfMissing(argv)
   var header = '# Change Log\n\nAll notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.\n'
   var oldContent = fs.readFileSync(argv.infile, 'utf-8')
@@ -150,34 +149,32 @@ function outputChangelog(argv, cb) {
   })
 }
 
-function getNextVersion(release) {
+function getNextVersion (release) {
   var newVersion = semver.inc(pkg.version, release.releaseAs)
 
   if (argv.preRelease) {
     // hack: if a feat is commited after two patches it wants to increase
     // the prerelease tag instead of the minor and start from 0
     // the next 2 lines and the if statement should fix this
-    realVersion = semver.inc(pkg.version, release.releaseAs)
-    prereleaseVersion = semver.inc(pkg.version, "prerelease", argv.tag)
+    var realVersion = semver.inc(pkg.version, release.releaseAs)
+    var prereleaseVersion = semver.inc(pkg.version, 'prerelease', argv.tag)
 
     if (semver.gt(realVersion, prereleaseVersion)) {
-
       newVersion = semver.inc(pkg.version, release.releaseAs)
-      check = newVersion
-      newVersion = newVersion + "-" + argv.tag + ".0"
+      newVersion = newVersion + '-' + argv.tag + '.0'
 
       if (semver.gt(prereleaseVersion, newVersion)) {
         newVersion = prereleaseVersion
       }
     } else {
-      newVersion = semver.inc(newVersion, "prerelease", argv.tag)
+      newVersion = semver.inc(newVersion, 'prerelease', argv.tag)
       console.log('normal: ' + newVersion)
     }
   }
   return newVersion
 }
 
-function commit(argv, newVersion, cb) {
+function commit (argv, newVersion, cb) {
   var msg = 'committing %s'
   var args = [argv.infile]
   var verify = argv.verify === false || argv.n ? '--no-verify ' : ''
@@ -187,7 +184,7 @@ function commit(argv, newVersion, cb) {
   }
   checkpoint(msg, args)
 
-  function handleExecError(err, stderr) {
+  function handleExecError (err, stderr) {
     // If exec returns an error or content in stderr, log it and exit with return code 1
     var errMessage = stderr || (err && err.message)
     if (errMessage) {
@@ -205,11 +202,11 @@ function commit(argv, newVersion, cb) {
   })
 }
 
-function formatCommitMessage(msg, newVersion) {
+function formatCommitMessage (msg, newVersion) {
   return String(msg).indexOf('%s') !== -1 ? util.format(msg, newVersion) : msg
 }
 
-function tag(newVersion, argv) {
+function tag (newVersion, argv) {
   var tagOption
   if (argv.sign) {
     tagOption = '-s '
@@ -232,7 +229,7 @@ function tag(newVersion, argv) {
   })
 }
 
-function createIfMissing(argv) {
+function createIfMissing (argv) {
   try {
     accessSync(argv.infile, fs.F_OK)
   } catch (err) {
@@ -244,8 +241,8 @@ function createIfMissing(argv) {
   }
 }
 
-function checkpoint(msg, args, figure) {
+function checkpoint (msg, args, figure) {
   console.info((figure || chalk.green(figures.tick)) + ' ' + util.format.apply(util, [msg].concat(args.map(function (arg) {
-      return chalk.bold(arg)
-    }))))
+    return chalk.bold(arg)
+  }))))
 };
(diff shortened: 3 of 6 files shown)


index.js
```diff
@@ -38,6 +38,19 @@ var argv = require('yargs')
...
     global: true
   })
+  .option('pre-release', {
+    alias: 'p',
+    describe: 'Should the release be a pre-release?',
+    type: 'boolean',
+    default: false,
+    global: true
+  })
+  .option('pre-id', {
```
stevemao: This should be call "tag", as https://docs.npmjs.com/cli/publish

Also you need to change the final `npm publish` information

stevemao:
I wonder if we could leverage `unreleased` option in conventional-changelog.

.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 .nyc_output
+.idea
```
stevemao: Don't do unrelated changes :)

index.js
stvnwrgs: Defining a tag when publishing is the same like a git tag. pre-release and pre-id are one part of the tag.  See http://semver.org/ for more infos.

stevemao: Where did you see the name "pre-id"?



> About the publish, yes but this is in the users scope.

What I mean is just to display a recommendation. Because right now it says [`npm publish` without a tag](https://github.com/conventional-changelog/standard-version/blob/master/index.js#L158-L160). Displaying `--tag` would be a better recommendation.

stevemao: I'm not completely sure about the relationship between git tag and npm publish tag. It doesn't look the same to me. npm publish by default uses "latest" tag.

stvnwrgs: Sorry, @stevemao you are right. The npm publish tag is not the same like the git tag.

stvnwrgs:
**After using it for a while, I found two bugs that need's to be fixed before this can be merged.**

One is nearly fixed, the second one needs some discussion.

Should prereleases be shown in the changelog?
I think only normal releases should be added by default.
If so, that would might lead to empty changelogs entries for normal versions and the prereleases are spamming the changelog.

What do you think about the following:
Changelogs are only written for non prereleases by default. An optional parameter would be possible to enable changelogs for prereleases. 

For this we could only write a changelog if the version is a normal one.
I already tried this, but it looks like conventional-changelog-core is then not finding all changes since the last normal release. Can you explain a little bit, how conventional-changelog-core is setting the delta from the newest and last written commit?

stevemao:
What bugs? Can you remove all unrelated changes and separate different things into different issues?

stvnwrgs:
The bus are fixed with the two latest commits. If using a prerelease versions new features did not increase the minor version, instead they just increased the prerelease, which was wrong.

The only unrelated thing right now is the next, actual feature. I could remove it, wait until the PR is merged and readd it, but its just 15 lines of code. In my opinion its not worth the time. But if you insist, I will do it.

Should I remove the changelog entry?

stevemao:
I'd suggest you to create another PR for other things. You also need to add tests. Also too many unnecessary whitespace changes.

stvnwrgs:
@stevemao: What about the package.json and CHANGELOG.md? I think it does not make sense to generate them in the PR, right?

stevemao:
Yeah, as atomic as possible :)

stvnwrgs:
Should be ready now. Thanks for your help!

index.js
```diff
@@ -62,22 +76,27 @@ conventionalRecommendedBump({
...
+    newVersion = getNextVersion(release)
     checkpoint('bumping version in package.json from %s to %s', [pkg.version, newVersion])
     pkg.version = newVersion
     fs.writeFileSync(pkgPath, JSON.stringify(pkg, null, 2) + '\n', 'utf-8')
   } else {
     checkpoint('skip version bump on first release', [], chalk.red(figures.cross))
   }
-
-  outputChangelog(argv, function () {
+  // only commit the package.json on prerelease
```
stevemao: Why?

test.js
```diff
@@ -198,3 +257,4 @@ describe('cli', function () {
     execCli('-n').code.should.equal(0)
   })
 })
+
```
stevemao: unnecessary whitespace, will be stripped by editors like atom by default.

index.js
```diff
@@ -38,6 +38,20 @@ var argv = require('yargs')
     default: false,
     global: true
   })
+  .option('pre-release', {
```
stevemao: We can probably merge this option with `tag`?

index.js
```diff
@@ -38,6 +38,20 @@ var argv = require('yargs')
...
+    alias: 'p',
+    describe: 'Should the release be a pre-release?',
+    type: 'boolean',
+    default: false,
+    global: true
+  })
+  .option('tag', {
+    alias: 't',
+    describe: 'Your pre-release id: e.g. beta, dev',
```
stevemao: Your pre-release tag

README.md
```diff
@@ -105,6 +105,20 @@ npm run release -- --no-verify
 standard-version --no-verify
 ```
 
+### Create a Prerelease
+```sh
+# identifier: alpha, beta, etc
+standard-version -p --tag="<identifier>"
+```
+
+### Dry run
```
stevemao: Remove this for this PR?

index.js
```diff
@@ -38,6 +38,20 @@ var argv = require('yargs')
...
     global: true
   })
+  .option('pre-release', {
+    alias: 'p',
+    describe: 'Should the release be a pre-release?',
+    type: 'boolean',
+    default: false,
+    global: true
+  })
+  .option('tag', {
```
stevemao: Sorry, I think this should be called "Prerelease Identifier". I'm bad at naming so just my own opinion...

index.js
```diff
@@ -108,6 +127,31 @@ function outputChangelog (argv, cb) {
   })
 }
 
+function getNextVersion (release) {
+  var newVersion = semver.inc(pkg.version, release.releaseAs)
+
+  if (argv.preRelease) {
```
stevemao: hmm.. I think you could just do something like `semver.inc(pkg.version, 'premajor', argv.tag)` if it's a major bump.

stevemao:
I wonder if we should let `conventional-recommended-bump` to decide if this is a prerelease?

stevemao:
There is also a from-git option in [npm-version](https://docs.npmjs.com/cli/version)

README.md
stvnwrgs: Why should documentation for new features of a PR not be included?

index.js
stvnwrgs: So the version for a prerelease is bumped, but no docs are generated. This only happens when the release is stable. It would flood the changelog otherwise.

index.js
stvnwrgs: I think I did this in a previous commit. But then no new minor/major version is created if feat/ breacking change comes. If you find a better way for this problem I'm open for it!

stvnwrgs:
Really, after 4 days you come up `conventional-recommended-bump` ?

index.js
stevemao: See https://github.com/conventional-changelog/conventional-changelog/blob/master/CHANGELOG.md#010-beta3-2015-07-17 it works well for me

stevemao:
What do you mean? I'm just saying if `conventional-recommended-bump` can detect prereleases

README.md
stevemao: The code for this is not in this PR yet

index.js
bcoe: @stvnwrgs, I think I'm with @stevemao. Let's generate a CHANGELOG for pre-releases, with this caveat:

It would be neat if there was eventually logic that could once `v1.0.0` is released, combine the information from  `v1.0.2.alpha.0`, `v1.0.3.alpha.1` into a single `v1.0.0` entry in the CHANGELOG; I would start simple though and not have a special case for pre-release handling.

index.js
```diff
@@ -108,6 +127,31 @@ function outputChangelog (argv, cb) {
...
+    if (semver.gt(realVersion, prereleaseVersion)) {
+      newVersion = semver.inc(pkg.version, release.releaseAs)
+      newVersion = newVersion + '-' + argv.tag + '.0'
+
+      if (semver.gt(prereleaseVersion, newVersion)) {
+        newVersion = prereleaseVersion
+      }
+    } else {
+      newVersion = semver.inc(newVersion, 'prerelease', argv.tag)
+      console.log('normal: ' + newVersion)
```
bcoe: let's drop this log line.

index.js
```diff
@@ -108,6 +127,31 @@ function outputChangelog (argv, cb) {
...
+    // the prerelease tag instead of the minor and start from 0
+    // the next 2 lines and the if statement should fix this
+    var realVersion = semver.inc(pkg.version, release.releaseAs)
+    var prereleaseVersion = semver.inc(pkg.version, 'prerelease', argv.tag)
+
+    if (semver.gt(realVersion, prereleaseVersion)) {
+      newVersion = semver.inc(pkg.version, release.releaseAs)
+      newVersion = newVersion + '-' + argv.tag + '.0'
+
+      if (semver.gt(prereleaseVersion, newVersion)) {
```
bcoe: I'm missing why this second check is necessary, mind walking me through the state machine of this logic a tiny bit.

bcoe:
@stvnwrgs this is awesome work, would love to work with you to get it merged; I've been balancing `standard-version` with a few other OSS projects, and apologize that the turn around time on our issues can sometimes be moderately frustrating.

index.js
stvnwrgs: @bcoe: Yes, this is really a little bit tricky here. It is also really hard to explain, but let me try.

This is a little bit hacky =(, because semver is not acting like expected. If you are at version 1.0.0 and commit(fix) and create a prerelease, semver gives you 1.0.1-beta.0, which is right.

If you add another commit now with a feature, and you create another prerelease it should become 1.1.0-beta.0. The problem is, semver gives you 1.0.1-beta.1 back.

So, line 140 is checking for this. If semver originally wants to create a bigger version (major or minor instead of patch) we want to use the  newVersion and append the tag + build it 0.

The second line is to ensure, that a new commit would result in an increase of the build id.

When I just see the code, id does not really make sense for me too. I removed the lines and tested it. Maybe this could be rewritten somehow for better understanding? Tests are checking this.

bcoe: @stvnwrgs rather than a rewrite, mind adding a test for each of the branches in the logic.

If any of the branches aren't necessary, this will probably jump out at you as you write the tests.

index.js
Tapppi: I'm with @bcoe, the end goal should be combining them to a single entry eventually. Another option is to collect the changes in a `Unrelease vX.Y.Z` and then name it to `vX.Y.Z` once it's not a pre-release anymore. I could swear I've seen that around, although I'm too lazy now to search for it.

Tapppi:
This feature would be awesome. @stvnwrgs do you have interest in continuing work on this, or do you prefer someone else continue if they have time? I intend to review this more thoroughly as soon as I have a slot of time open.

AntJanus:
@Tapppi @stvnwrgs I'd be really interested in seeing this feature pass. What else is left? Maybe I can fork + contribute

stevemao:
@AntJanus you are more than welcome to pick up the work 😄  open source projects are about community 😃

stvnwrgs:
@AntJanus I'm out. Feel free!

index.js
Tapppi: @bcoe I thought about combining the pre-release logs to a single one, but what about pre-releases containing fixes to other pre-releases. e.g. You do a pre-release of a new feature as `v2.1.0-rc.0` with `feat(god): can now obtain god powers`, but then fix an edge case of this new feature in the next prerelease `v2.1.0-rc.1`: `fix(god): fix edge case to give full god instead of demigod powers`. Both of them end up in the combined `v2.1.0` changelog and it might end up with quite a lot of 'fixes inside of this release' that are out of context.

bcoe:
@AntJanus I'd love to see this feature too, perhaps you can start by opening an issue; and we can discus what we'd like to see out of pre-release functionality? Then we can help see @stvnwrgs hard work over the finish line.

AntJanus:
@bcoe Opened an issue on #122

index.js
```diff
@@ -108,6 +127,31 @@ function outputChangelog (argv, cb) {
...
+        newVersion = prereleaseVersion
+      }
+    } else {
+      newVersion = semver.inc(newVersion, 'prerelease', argv.tag)
```
e-cloud: the third argument of `inc()` is not for identifier

bcoe:
closing in favor of https://github.com/conventional-changelog/standard-version/pull/129 which looks almost ready to land.

PR #103, opened 2016-09-12
fix: check private field in package.json(#102)

### LINKED ISSUE #102: Should not suggest `npm publish` if the package is private
Hello! I'm using `standard-version` in my private package, and the private package never is published. So my request is if the `package.json` has `private:true`, then the result doesn't display the `npm publish`.
```diff
$ standard-version patch
✔ bumping version in package.json from 1.0.4 to 1.0.5
✔ outputting changes to CHANGELOG.md
✔ committing package.json and CHANGELOG.md
✔ tagging release 1.0.5
++ ℹ Run `git push --follow-tags origin master` to publish
-- ℹ Run `git push --follow-tags origin master; npm publish` to publish
```

Thanks.

watilde:
Fix #102. It should not suggest `npm publish` if the package.json has `private: true`.


nexdrew:
LGTM, thanks @watilde!

Issue #80, opened 2016-08-02
Review before commit


dignifiedquire:
It would be very nice if there was a way to review 
- the suggested version and 
- the changelog that was generated
  during the workflow and adjust them if needed. Currently I have to go back and squash fix commits or redo to whole thing. 

I imagine for the version it could be a nice interactive prompt telling you the version it will use and ask for either confirmation or a different version to be entered.

For the changelog it could work like `git commit` works, in that it opens the default `$EDITOR` and as soon as the file there is saved and closed it drops back into continuing the work.

stvnwrgs:
Yeah, had the same issues, directly added it in #83. Sorry, that it's not in a seperate PR.

bcoe:
@dignifiedquire sorry that it took so long! but this is now implemented with the flag `--dry-run`.

PR #106, opened 2016-09-15
feat: dry run mode

azachar:
Hi there, 
I implemented a dry run feature (--test switcher) that doesn't bump anything but instead it just shows the version of the next release.

It comes with the tests too.

Cheers,
Andrej

### DIFF
--- a/index.js
+++ b/index.js
@@ -38,6 +38,13 @@ var argv = require('yargs')
     default: false,
     global: true
   })
+  .option('test', {
+    alias: 't',
+    describe: 'Dry run, display a new next version',
+    type: 'boolean',
+    default: false,
+    global: true
+  })
   .help()
   .alias('help', 'h')
   .example('$0', 'Update changelog and tag release')
@@ -66,11 +73,20 @@ conventionalRecommendedBump({
   var newVersion = pkg.version
   if (!argv.firstRelease) {
     newVersion = semver.inc(pkg.version, release.releaseType)
+
+    if (argv.test) {
+      checkpoint('A new version will be from %s to %s', [pkg.version, newVersion])
+      return
+    }
+
     checkpoint('bumping version in package.json from %s to %s', [pkg.version, newVersion])
     pkg.version = newVersion
     fs.writeFileSync(pkgPath, JSON.stringify(pkg, null, 2) + '\n', 'utf-8')
   } else {
     checkpoint('skip version bump on first release', [], chalk.red(figures.cross))
+    if (argv.test) {
+      return
+    }
   }
 
   outputChangelog(argv, function () {
--- a/test.js
+++ b/test.js
@@ -88,6 +88,32 @@ describe('cli', function () {
       content.should.match(/1\.0\.1/)
       content.should.not.match(/legacy header format/)
     })
+
+    it('should not make any changes with the test option', function () {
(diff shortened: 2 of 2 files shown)


e-cloud:
why not use dry-run as argument directly?

Tapppi:
This is a good idea, but I think this and #80 are the same problem with two different solutions. no? IMO a preview of the changes to both the version and CHANGELOG.md is preferable to just a dry-run. Do you think a preview would solve your problem @azachar?

/CC @bcoe @stevemao

azachar:
Yep, preview is also fine, as soon as it keeps telling in the console which version it is going the be bumped. I really do not want to switch to ide just to check that. 

 Btw. since there is no PR for #80, I suggest that this can be added later to this PR. Calling feature ---test, or --preview or --dry-run it all sounds ok to me. 

Cheers,
A.

index.js
```diff
@@ -38,6 +38,13 @@ var argv = require('yargs')
     default: false,
     global: true
   })
+  .option('test', {
```
bcoe: let's go with `d` and `--dry-run`.

index.js
```diff
@@ -66,11 +73,20 @@ conventionalRecommendedBump({
   var newVersion = pkg.version
   if (!argv.firstRelease) {
     newVersion = semver.inc(pkg.version, release.releaseType)
+
+    if (argv.test) {
```
bcoe: rather than having special logic that prints this message; I'd prefer that we build `--dry-run` such that it runs an identical code path, but mocks the git and fs logic It would also be very cool if we used something like `chalk` to print some pretty output to the terminal, indicating what will go in the CHANGELOG.md.

test.js
```diff
@@ -189,6 +215,18 @@ describe('cli', function () {
     pkgJson.should.equal(['{', '  "version": "1.0.1"', '}', ''].join('\n'))
   })
 
+  it('should not make any changes with the test option', function () {
```
bcoe: appreciate the tests!

index.js
azachar: yep, let's do it.  Should I update my PR?

test.js
azachar: You are welcome :) I love standards and making tests is one of them :)

index.js
azachar: Yep, feel free to extend my PR in that way, it all sound good to me!

joefraley:
what's the status on this discurrion? seems very useful

bcoe:
@joefraley I love the idea of this feature -- perhaps we should take another stab at it?

joefraley:
Sure, what kept it from merging before?

bcoe:
@joefraley mainly I'd like the command `--dry-run` to show every step that would take place:

the commit, the CHANGELOG that would be generated, etc; it might be a good opportunity to refactor a little bit so that steps are a distinct `class` -- I'm looking to the future and thinking about how we'll handle rollbacks.

bcoe:
@joefraley @azachar @Tapppi I've overhauled standard-version a bit, and split execution into [several clearly defined steps](https://github.com/conventional-changelog/standard-version#lifecycle-scripts), during this work I implemented `--dry-run` with a preview.

Would love your help testing it out, before it gets released to a wider audience:

`npm i standard-version@next`

Issue #91, opened 2016-08-17
Exit only on error code?


duclet:
So I have a post-commit hook which actually undo a commit if the history is not fully linear. Basically, it seems like any output that is made thru a commit hook, will end up on stderr. Since the code checks against that, it never gets to the tagging.

stevemao:
> it seems like any output that is made thru a commit hook,

look like a bug in the commit hook

Tapppi:
Idd, what are you using to create the commit-hook?

duclet:
The problem isn't the hook itself. It seems like to me, all output from programs executed by the git hook forcefully redirect standard out to standard err. I've verify this by running the program directly and it outputs to stdout. When running from the githook, it redirects to stderr.

Perhaps my question is also on why it was decided to terminate execution when there is a message on stderr even though the exit code signals it is okay? In terms out logging and outputting messages, many programs I see can emit warnings (which generally goes to stderr), not just errors. When there is an error, it should be terminating with an exit code greater than 0.

e-cloud:
[Source code here](https://github.com/conventional-changelog/standard-version/blob/master/index.js#L123)

One solution, when there is `null` err, and has stderr, the message may output as warnning and the program should not exit.

nexdrew:
I agree with @duclet and @e-cloud - `standard-version` should not fail-fast simply from stderr output. I like the suggestion to "output a warning and continue". There's no reason to fail if git exits 0 (without an error).

PR #110, opened 2016-09-22
Don't fail on stderr output, but print the output to stderr

Tapppi:
For #91, this would stop `standard-version` from failing fast when `git` or hooks or something prints to `stderr`, but prints it out. Added test case it for as well. Also, errors are now printed to `stderr/console.error` instead of `stdout/console.log`.

/CC @nexdrew @duclet @e-cloud @stevemao @bcoe

### DIFF
--- a/index.js
+++ b/index.js
@@ -108,6 +108,17 @@ function outputChangelog (argv, cb) {
   })
 }
 
+function handleExecError (err, stderr) {
+  // If exec returns an error or content in stderr, log it and exit with return code 1
+  var errMessage = stderr || (err && err.message)
+  if (errMessage) {
+    console.error(chalk.red(errMessage))
+    if (err) {
+      process.exit(1)
+    }
+  }
+}
+
 function commit (argv, newVersion, cb) {
   var msg = 'committing %s'
   var args = [argv.infile]
@@ -118,14 +129,6 @@ function commit (argv, newVersion, cb) {
   }
   checkpoint(msg, args)
 
-  function handleExecError (err, stderr) {
-    // If exec returns an error or content in stderr, log it and exit with return code 1
-    var errMessage = stderr || (err && err.message)
-    if (errMessage) {
-      console.log(chalk.red(errMessage))
-      process.exit(1)
-    }
-  }
   exec('git add package.json ' + argv.infile, function (err, stdout, stderr) {
     handleExecError(err, stderr)
     exec('git commit ' + verify + (argv.sign ? '-S ' : '') + 'package.json ' + argv.infile + ' -m "' + formatCommitMessage(argv.message, newVersion) + '"', function (err, stdout, stderr) {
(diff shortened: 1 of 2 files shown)


Tapppi:
I suspect coverage decrease is because the actual amount of code went down, since the `git tag`s error handling got rolled into `handleExecError`? Can't figure out another reason.

index.js
```diff
@@ -108,6 +108,17 @@ function outputChangelog (argv, cb) {
   })
 }
 
+function handleExecError (err, stderr) {
```
stevemao: I thought we should handle `err` and `stderr` differently?
the problem is with `stderr` shouldn't terminate the program. But for `err` we keep the same behaviour. No?

stevemao: I think I misread your code. It LGTM now.

stevemao:
> I suspect coverage decrease is because the actual amount of code went down, since the git tags error handling got rolled into handleExecError? Can't figure out another reason.

You can check out the coverage output and see why

index.js
```diff
@@ -108,6 +108,17 @@ function outputChangelog (argv, cb) {
...
+  // If exec returns an error or content in stderr, log it and exit with return code 1
+  var errMessage = stderr || (err && err.message)
+  if (errMessage) {
+    console.error(chalk.red(errMessage))
```
e-cloud: i think when `err` is null, better to use `console.warn` to output the error message. Red message means real error.
Since console.error/warn's output will have its own color, should it continue to use `chalk` to control the color?

stevemao: > Since console.error/warn's output will have its own color

I don't think so

Tapppi: > The console.warn() function is an alias for console.error().
https://nodejs.org/api/console.html#console_console_warn_data

Tapppi: I think we should print to `stderr` (`console.error`), but how about yellow color for when the exit code was 0?

Tapppi:
@stevemao:
> Is it possible to test exit code for all different scenarios?

What do you mean by this? What scenarios should we be checking for? One thing that is missing, is rolling back changes made by `standard-version` when the whole process doesn't go through, but I think that's another PR as it's a bigger change.

> You can check out the coverage output and see why

Thanks! Yeah, the missed line count stayed the same at 4, but overall line count went down. Should actually find out what is not being tested.. I'll get to it soon, as I'll have some time.

index.js
e-cloud: sorry for my poor knowledge, yellow would be great!

Tapppi: Done in 458da35

stevemao:
> Is it possible to test exit code for all different scenarios?

looks like we only test `result.code.should.equal(0)`, but not `1`.  As long as we cover all scenarios we are fine :)

Nice work! @Tapppi

index.js
```diff
@@ -109,10 +109,11 @@ function outputChangelog (argv, cb) {
...
-  // If exec returns an error or content in stderr, log it and exit with return code 1
+  // If exec returns content in stderr, but no error, print it as a warning
+  // If exec returns an error, print it and exit with return code 1
   var errMessage = stderr || (err && err.message)
   if (errMessage) {
-    console.error(chalk.red(errMessage))
+    console.error(err ? chalk.red(errMessage) : chalk.yellow(errMessage))
     if (err) {
       process.exit(1)
     }
```
e-cloud: @Tapppi , i think it would be better to be:
```js
    if (err) {
      console.error(chalk.red(stderr || err.message))
      process.exit(1)
    } else if(stderr) {
      console.warn(chalk.yellow(stderr))
    }
```
the prior `stderr` in `console.error` could be dropped.

Tapppi: Good idea, changed in 6338f32

bcoe:
@Tapppi looks good to me land it.

e-cloud:
when will this fix release? coz now it's some kind of blocking bug.

e-cloud:
a beta release is welcomed too.

stevemao:
@bcoe could you publish this?

PR #113, opened 2016-09-24
chore: add os, npm and editor related ignores to git

Tapppi:


stevemao:
I put a lot of them in the global gitignore. CC @bcoe @nexdrew

PR #114, opened 2016-09-24
Restructuring for 100% test coverage

Tapppi:
This might be a bit much, but actually nice in the long run. So the only parts that weren't being tested previously were `conventionalChangelog` and `conventionalRecommendedBump` error handling. The problem with handling them was that the `cli` was being called with shell execution, so mocking was nigh impossible. As a result, I did some restructuring:

* Separated the `cli` into `cli.js`
* `standard-version` actual logic in `index.js` with `module.exports` being a function
    * This allows `standard-version` to be used by other tools, tests etc.
    * Usage: `require('standard-version')(opts, callback)`
    * `opts` is an object with `cli` options in camelcase, e.g. `{firstRelease: true}`
    * `callback` will be called with possible error when `standard-version` is done (see `cli.js`)
* Separated default options into `defaults.json`, so they can be at one place instead of both `index.js` and `cli.js` (needed in `cli.js` for `yargs` help messages)
* Wrote tests for `conventionalChangelog` and `conventionalRecommendedBump` error situations
    * Mocked with [`mockery`](https://github.com/mfncooper/mockery)
* Added `silent` option to skip logging to console, also added to the `cli` (needed in tests that use `index.js` directly)

Long list of small changes all related, needed to get test coverage to 100%. I think the structure is nice, but the interface could probably actually be an event emitter or generator that emits events at checkpoints. Then the `cli` could log those checkpoints and the silent option wouldn't actually be needed in `index.js`. Just an idea, but that would separate logic that's always wanted and logic that's `cli` specific.

I'll work on working these into sensible commits with messages once theres some discussion about it, but that shouldn't take very long.

/CC @bcoe @stevemao @nexdrew

### DIFF
--- a/index.js
+++ b/index.js
@@ -143,16 +143,20 @@ function createIfMissing (argv) {
 }
 
 function checkpoint (argv, msg, args, figure) {
-  console.info((figure || chalk.green(figures.tick)) + ' ' + util.format.apply(util, [msg].concat(args.map(function (arg) {
-    return chalk.bold(arg)
-  }))))
+  if (!argv.silent) {
+    console.info((figure || chalk.green(figures.tick)) + ' ' + util.format.apply(util, [msg].concat(args.map(function (arg) {
+      return chalk.bold(arg)
+    }))))
+  }
 }
 
 function printError (argv, msg, opts) {
-  opts = objectAssign({
-    level: 'error',
-    color: 'red'
-  }, opts)
+  if (!argv.silent) {
+    opts = objectAssign({
+      level: 'error',
+      color: 'red'
+    }, opts)
 
-  console[opts.level](chalk[opts.color](msg))
+    console[opts.level](chalk[opts.color](msg))
+  }
 }
--- a/index.js
+++ b/index.js
@@ -1,84 +1,54 @@
-#!/usr/bin/env node
 var conventionalRecommendedBump = require('conventional-recommended-bump')
 var conventionalChangelog = require('conventional-changelog')
 var path = require('path')
-var argv = require('yargs')
-  .usage('Usage: $0 [options]')
-  .option('infile', {
-    alias: 'i',
-    describe: 'Read the CHANGELOG from this file',
-    default: 'CHANGELOG.md',
-    global: true
-  })
-  .option('message', {
-    alias: 'm',
-    describe: 'Commit message, replaces %s with new version',
-    type: 'string',
-    default: 'chore(release): %s',
-    global: true
-  })
-  .option('first-release', {
-    alias: 'f',
-    describe: 'Is this the first release?',
-    type: 'boolean',
-    default: false,
-    global: true
-  })
-  .option('sign', {
-    alias: 's',
-    describe: 'Should the git commit and tag be signed?',
-    type: 'boolean',
-    default: false,
-    global: true
-  })
-  .option('no-verify', {
-    alias: 'n',
-    describe: 'Bypass pre-commit or commit-msg git hooks during the commit phase',
-    type: 'boolean',
-    default: false,
-    global: true
-  })
-  .help()
-  .alias('help', 'h')
-  .example('$0', 'Update changelog and tag release')
-  .example('$0 -m "%s: see changelog for details"', 'Update changelog and tag release with custom commit message')
-  .wrap(97)
-  .argv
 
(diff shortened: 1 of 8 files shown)


Tapppi:
Still getting timeouts in travis it seems.. Should we increase it to 30000ms possibly. Travis is ridiculously slow though.

index.js
```diff
@@ -1,181 +1,159 @@
...
+  }, function (err, release) {
+    if (err) {
+      printError(err.message)
+      return done(err)
+    }
 
-  changelogStream.on('data', function (buffer) {
-    content += buffer.toString()
-  })
+    var newVersion = pkg.version
```
bcoe: Love this refactor! I'd be tempted to pull the methods in `standardVersion` out to the top level:

`module.exports = function standardVersion()`

`outputChangelog()`
`commit()`
...

test.js
```diff
@@ -141,20 +135,69 @@ describe('cli', function () {
...
+  describe('with mocked conventionalRecommendedBump', function () {
+    beforeEach(function () {
+      mockery.enable({warnOnUnregistered: false, useCleanCache: true})
+      mockery.registerMock('conventional-recommended-bump', function (_, cb) {
+        cb(new Error('bump err'))
+      })
+    })
+
+    afterEach(function () {
+      mockery.deregisterMock('conventional-recommended-bump')
```
bcoe: love the use of `mockery`, a new library I should play with 👍

Tapppi: Yeah I've been needing something like `mockery` for a while, and I cant believe I missed it before when I was looking for it.. It's really cool.

index.js
Tapppi: Yeah I first structured it like that, but will have to make sure everything is passed around correctly. I'll try to work that in before landing this.

Tapppi: @bcoe, this is now done in 440a675

Tapppi:
So since this seems to be viewed as the right way forward, I rebased and refactored:

* Now only the base function is in the main export (`standardVersion`). Everything else is a top level function
* As a result of the above, `argv` and `callback` passing around is consistent all around now.
* `handleExecError` is now `handledExec` to simplify error handling. It expects a different callback for errors and success
*  Tests using `index.js` directly have been refactored into their own `describe()` block (with one shared test to make sure the basic functionality works when `standardVersion` is used directly)

bcoe:
:shipit:

Tapppi:
This has now been refactored into sensible commits with messages that describe the changes done. The only actual change from before is that the `silent` flag is now known as `quiet` ( `-q`) because the alias `-s` was conflicting with the `sign` flag's alias.. :shipit: in progress

Tapppi:
`node 0.10` prints a ton of warnings from deps that want `>0.12` or `>4.0` and also prints something weird on `0.10` with the `--quiet` flag passed.. It's breaking the `--quiet` test. Can we just drop `0.10` and `0.12` here, since they would be dropped in #81 anyway? 

Also, now that I think of it, `--quiet` sounds like it would print errors but not normal output.. Is there a possibility that we'll want it to do that? Because in that case i think `--silent` without an alias is better than `--quiet` for squelching all output and `quiet` can be reserved for something else. Thoughts @bcoe @stevemao @nexdrew ? If not, I'll put all of those changes in tomorrow and merge this.

bcoe:
@Tapppi the underlying problem is a dependency of a dependency:

```js
const gitconfig = require('gitconfiglocal');
^^^^^
SyntaxError: Use of const in strict mode.
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at repl:1:2
    at REPLServer.self.eval (repl.js:110:21)
    at Interface.<anonymous> (repl.js:239:12)
    at Interface.emit (events.js:95:17)
```

I say let's drop `0.10` and `0.12` support 👍

bcoe:
make sure that this is a breaking change when you land it though:

`BREAKING CHANGE: dropping support for Node < 4`

stevemao:
Awesome work @Tapppi !

bcoe:
@seconded awesome work @Tapppi, this is now available in:

`+ standard-version@3.0.0`

PR #127, opened 2016-10-07
docs: Document commit-all flag with example

### LINKED ISSUE #96: Ability to add other artifacts to the release commit
ATM, I'm amending because I want the produced artifacts to be part of the `chore(release): x.x.x` commit.
```json
"prerelease": "webpack -p --bail", //produces some bundle artifacts that I want to commit
"release": "standard-version -s",
"postrelease": "git commit -a --amend --no-edit"
```

However, [because of the way git works](http://stackoverflow.com/a/3143658/1444541) this creates a new commit object since the original commit is tagged _before_ the amend happens. What would be nice is if my `postrelease` script was really a `poststandard-version` script (i.e. a "lifecycle-style" hook like the npm run script itself).

cc @stevemao @bcoe

JaKXz:
Resolves #96.

@Tapppi @bcoe please take a look :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,14 @@ standard-version --no-verify
 
 If you have your GPG key set up, add the `--sign` or `-s` flag to your `standard-version` command.
 
+### Committing generated artifacts in the release commit
+
+If you want to commit generated artifacts in the release commit (e.g. [#96](https://github.com/conventional-changelog/standard-version/issues/96)), you can use the `--commit-all` or `-a` flag. You will need to stage the artifacts you want to commit, so your `release` command could look like this:
+
+```json
+"release": "git add <file(s) to commit> && standard-version -a"
+```
+
 ### CLI Help
 
 ```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -111,6 +111,14 @@ standard-version --no-verify
 
 If you have your GPG key set up, add the `--sign` or `-s` flag to your `standard-version` command.
 
+### Committing generated artifacts in the release commit
+
+If you want to commit generated artifacts in the release commit (e.g. [#96](https://github.com/conventional-changelog/standard-version/issues/96)), you can use the `--commit-all` or `-a` flag. You will need to stage the artifacts you want to commit, so your `release` command could look like this:
+
+```json
```
bcoe: might be good to provide an example of a command that might produce the artifacts, .e.g., 

`prerelease: "webpack -p --bail"`

Tapppi: I'm with @bcoe I would add the `prerelease` script for context. Otherwise this looks good. IMHO it's better to give that one more line of context when people are not that likely to follow a link to a lengthy conversation.

JaKXz: Will do! :)

Tapppi: Sure, I don't think there's any real reason not to. We can remove it if someone deems it "unstylish" 😄 . Thanks!

JaKXz:
@Tapppi ready to go!

Tapppi:
Merging this as @bcoe already approved it too.

PR #128, opened 2016-10-08
refactor: use object-assign in test

JaKXz:
since util._extend is deprecated

### DIFF
--- a/test.js
+++ b/test.js
@@ -2,7 +2,7 @@
 
 'use strict'
 
-var extend = require('object-assign')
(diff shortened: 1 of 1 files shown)


test.js
```diff
@@ -2,7 +2,7 @@
 
 'use strict'
 
-var extend = Object.assign || require('util')._extend
+var extend = require('object-assign')
```
stevemao: Can you refactor the variable to `objectAssign`? Thanks :)

JaKXz: done

PR #129, opened 2016-10-15
support releasing a custom version, including pre-releases

e-cloud:
## major changes
- add a option `--release-as` to release a custom version like npm version
- add a option `--prerelease` to name a pre-release or tag a release as prerelease without option value, it can be used with `--release-as`

this PR is similar to #83 , but this provide more version-control.

---

**PS**: some of the test code are borrowed from tests for #83

**PPS**: I'm bad at naming the option. Feel free to suggest new names for those options.


e-cloud:
@Tapppi all your comment can be worked out except this:

> Prefixing the bump with `pre` if `prerelase` flag is used, and supplying either the given or default `prerelease` tag to `semver.inc`.

That involves some problems:
* `yargs` can specify a default value for an option, but can't not set as undefined when not using that option in cli. Then we aren't able to see if user specify a `prerelease` flag with simple way
* when someone wants no tag id, with default tag-id, it would also be hard to implement the requirement

so, the default value for `prerelease` is `undefined` and when a use specify it with option value, its value would be `''`(empty string). If a user need a custom tag id, he/she should specify it explicitly for now.

package.json
```diff
@@ -2,12 +2,12 @@
   "name": "standard-version",
   "version": "3.0.0",
   "description": "replacement for `npm version` with automatic CHANGELOG generation",
-  "bin": "cli.js",
+  "bin": "bin/standard-version.js",
```
stevemao: the naming needs to be changed. `cli` should be the `bin`. So you'd need to rename the files to whatever you think makes more sense :)

e-cloud: but why?

e-cloud:
@stevemao do you have time to review this again?

test.js
```diff
@@ -185,6 +198,126 @@ describe('cli', function () {
...
+      fs.writeFileSync('CHANGELOG.md', 'legacy header format<a name="1.0.0">\n', 'utf-8')
+
+      commit('fix: first commit')
+      execCli('--release-as').code.should.above(0)
+    })
+
+    describe('release-types', function () {
+      var regularTypes = ['major', 'minor', 'patch']
+
+      regularTypes.forEach(function (type) {
```
stevemao: I wouldn't auto generate tests like this. I'd keep it [DAMP](http://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests).

stevemao:
it looks awesome In general. Thanks @e-cloud. We also need to add docs before we can merge :)

test.js
e-cloud: But in this case we dont add a bunch of fixture to avoid duplication. I think its readability is acceptable, just opinion. It you insist, i can expand it. I would be much longer.

e-cloud:
I just add docs in Readme.md. Anybody could review this?

test.js
stevemao: I'd probably just test one for each assertion. EG: say test `'major'` for "creates a major release" and `'patch'` for "creates a pre-patch release". Not a big deal :)

bcoe:
@e-cloud really appreciate this work; sorry for the slow turn around on this repo lately, going to work on seeing this over the finish line 👍

bcoe:
@stevemao @Tapppi sorry that I've been so AWOL the past couple months -- haven't been able to carve as much OSS time out as I would like to.

Please feel free to land any pull requests that meet your guidelines; I've added you both to the npm module so that I don't block you on publishing, and I very much trust your instincts with regards to features.

bcoe:
@e-cloud @Tapppi give this a spin when you have a moment:

I've landed these changes _in a pre-release_! how meta is that.

Tapppi:
@bcoe @e-cloud sorry to bump this, but this is fricking awesome, I just released my first pre-release of an internal company package automagically from our CI server after versioning with `standard-version@4.0.0-0`  🎉  😍 ! All the changes look pretty good, I don't think there are any blocking issues for v4?! :shipit:

bcoe:
@Tapppi if you're comfortable signing off on everything, I say release it! my two cents:

* I think your point about rolling back the operation if an error is a good thing for us to aspire to; but perhaps we can take this on as we start to abstract the checkpoint logic... maybe model after database migrations? have a rollback for each checkpoint.
* I love the pre-release functionality, my only beef is changelog management gets slightly weird if you release several pre-releases in a row -- but I think this is something we can work on in the future too.

want to cut a release @Tapppi you should have access to the module on npm now too.

Tapppi:
@bcoe I think both of those points are crucial but big enough to warrant thinking a bit more about them instead of trying to rush them in. I'll give releasing `v4.0.0` a shot right now, thanks! :+1:

PR #76, opened 2016-07-26
feat: support cordova versioning

azachar:
Hello, 
I have implemented Cordova versioning support. So it bumps versions to `package.json` and `config.xml` too.

This PR contains doc and tests.

Best regards,
Andrej

PR #134, opened 2016-10-27
Add: bower support

lexich:


stevemao:
See https://github.com/conventional-changelog/standard-version/pull/76#issuecomment-235442954

mikefisher84:
Would love to see this merged as this is the only thing blocking our team from using this.

lexich:
+1 I think that this code can be merged and after when better solution will come - cut off this 'spagetty' code. But functionality will be available at this moment.

PR #143, opened 2016-11-20
feat: make tagPrefix configurable

niieani:
At Aurelia we don't use any tag prefixes, our tags are just `1.0`, thus we couldn't use `standard-version` as is. Right now `standard-version` has a hardcoded `v` prefix for all the git tags for releases, i.e. `v1.0`, `v1.1`. This change makes it configurable so you can have no prefix or a custom prefix if you wish. I left the default `v` prefix, so this is a non-breaking change.

### DIFF
--- a/command.js
+++ b/command.js
@@ -66,7 +66,7 @@ module.exports = require('yargs')
   .option('tag-prefix', {
     alias: 't',
     describe: 'Set a custom prefix for the git tag to be created',
-    type: 'boolean',
+    type: 'string',
     default: defaults.tagPrefix,
     global: true
   })
--- a/command.js
+++ b/command.js
@@ -63,6 +63,13 @@ module.exports = require('yargs')
     default: defaults.silent,
     global: true
   })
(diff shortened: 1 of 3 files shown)


cli.js
```diff
@@ -51,6 +51,13 @@ var argv = require('yargs')
     default: defaults.silent,
     global: true
   })
+  .option('tag-prefix', {
+    alias: 't',
+    describe: 'Set a custom prefix for the git tag to be created',
+    type: 'boolean',
```
bcoe: If you switch the type to `string` rather than `boolean` I believe the tests should start passing.

niieani: Thanks for feedback! I didn't notice the test failed, sorry about that. Corrected now, but I see merge conflict. I'll see if I can rebase.

bcoe:
@niieani I've landed your changes in an upcoming release, would love an extra set of eyes on it. To test it out:

👍

niieani:
Many thanks! I'll have a go at it.

PR #148, opened 2016-11-26
feat: add support for bumping version # in bower.json

bcoe:
working from @lexich's foundation, this pull adds support for bumping version # in multiple config files in a generic way.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,10 +13,8 @@ var objectAssign = require('object-assign')
 
 module.exports = function standardVersion (argv, done) {
   var pkgPath = path.resolve(process.cwd(), './package.json')
-  var bowerPath = path.resolve(process.cwd(), './bower.json')
   var pkg = require(pkgPath)
   var defaults = require('./defaults')
-
   var args = objectAssign({}, defaults, argv)
 
   bumpVersion(args.releaseAs, function (err, release) {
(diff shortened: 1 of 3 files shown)


bcoe:
@lexich I've released support for `bower.json` in the upcoming release of `standard-version`; I would love the extra set of eyes/help testing:

lexich:
@bcoe nice. I think that the next step should be to adding support pluggins for different file formats.

index.js
```diff
@@ -52,6 +47,37 @@ module.exports = function standardVersion (argv, done) {
...
+        var filename = path.basename(configPath)
+        config.version = newVersion
+        fs.writeFileSync(configPath, JSON.stringify(config, null, 2) + '\n', 'utf-8')
+        checkpoint(args, 'bumping version in ' + filename + ' from %s to %s', [config.version, newVersion])
+        // flag any config files that we modify the version # for
+        // as having been updated.
+        configsToUpdate[configPath] = true
+      }
+    } catch (err) {
+      if (err.code !== 'ENOENT') console.warn(err.message)
```
Tapppi: @bcoe This PR is so cool, thanks for getting back into the project so strongly, I've been AWOL due to increased pressure at our startup but am trying to catch up too. 

I'm slightly concerned about the fact that any write errors will be ignored, e.g. if you have a `bower.json` you would presumably want to have it updated too. If the write fails, you'll just get a error but the git commits etc. will go through!

I haven't yet settled on an opinion on how this should be handled, but I think it's part of a bigger problem related to rolling back changes if an error occurs during the whole run..

e-cloud: rolling back changes will be a awesome feature

Issue #130, opened 2016-10-17
How would I make a hotfix/patch release to an older branch/release


fritzvd:
Hi,

Is it possible to be able to release from a fixes/hotfix/stable whatever branch? 

E.g. master is ahead at 4.3.1. But I'm looking to release a security patch to 2.0.1 from a `release2.0` branch for it to become 2.0.2.

How would I go about that in standard-version? Or is that scenario not supported?

Fritz

P.S. I've been looking through the issues, and the most similar ones to mine are those about named versions (#83, #84). But that's not what I mean.

stevemao:
You could switch to `release2.0` branch, do commits on the branch and release. It should generate the right changelog. Merging it back to master might be a bit tricky I think. You'd need to resolve the conflicts in CHANGELOG.md
linking: https://github.com/conventional-changelog/git-semver-tags/pull/10#issuecomment-254356199

Am I missing anything?

fritzvd:
Thanks @stevemao. That makes sense.

stevemao:
@fritzvd does it work for you? Is this resolved?

Tapppi:
This should work at the moment as is, right @fritzvd? I think the right workflow in this situation is to:

* make the fix on a separate branch e.g. `hotfix-crash-on-startup-23` 
* merge this to the earlier `release2.0` branch 
* run `standard-version` which should pick up only the tags on that branch, and release a correct `CHANGELOG.md` and version.
* if possible merge the hotfix branch to your other release (`master`/`release2.1`/`release3.0` etc.) or rebase and ff-merge and repeat the standard-version process.

This avoids conflict between changelogs etc. as the same hotfix is just applied to multiple branches of release. IMO that is the correct workflow and should be supported by `standard-version` explicitly (right now this seems to be pretty obtuse and implicitly supported). As explaine in https://github.com/conventional-changelog/git-semver-tags/pull/10#issuecomment-254356199 having that pull requests result as default in `standard-version` would break this workflow with no alternative I can see? As such I still am of the opinion that the functionality should be in `git-semver-tags` as an option and disabled here until other concerns such as this issue are resolved.

Unless there is a dissenting opinion I think this can be closed as soon as someone verifies that my assumption about this working right now is correct?

/CC @bcoe @nexdrew @stevemao

Tapppi:
BTW we should probably document the working workflow. so:

* release from `master` and if you want individual `release2`etc. branches
[...]

PR #152, opened 2016-12-12
docs(readme): add guidelines for release branches as per #130

fritzvd:
Hi,

I took some time to merge some of the comments for the README, mostly based on @Tapppi's comments. Sorry it took a week longer to find the time.

If something is unclear or inaccurate I'll gladly change it.

Cheers,

Fritz

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,29 @@ you will get version `1.1.0` rather than the auto generated version `1.0.1`.
 
 > **NOTE:** you can combine `--release-as` and `--prerelease` to generate a release. This is useful when publishing experimental feature(s).
 
+### Release on a seperate branch like `release2.0` or `hotfix-parsing-lists`
+
+Releasing on a seperate branch is slightly more elaborate. Say for example you have a development version on master and a 2.0 version branch that should remain stable(-ish).
+
+In the case of a release2.0 branch with a hotfix switch to a new branch:
+
+    git checkout release2.0 # if you're not already on that branch
+    git checkout -b hotfix-i18n-broken
+
+After you're done commit your work and **first** merge it to the main branch before running the release.
+
+    git checkout release2.0
+    git merge hotfix-i18n-broken # or rebase whichever you prefer
+
+    # npm run script
+    npm run release
+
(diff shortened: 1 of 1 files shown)


fritzvd:
Whoops sorry haven’t found the time to get back to this. I’ll probably have some more time next week

Fritz

+31 (0) 6 16 058 218
pgp public key: https://keybase.io/fritzvd/key.asc

bcoe:
@fritzvd awesome, thanks for adding this content; One thing that comes to my mind is I would love a better story for pre-releases;

I find releasing from `master` is rarely an issue once code is ready to go out the door; but I often find I want to release intermediate library versions from a branch (usually this release is related to a specific ticket):

https://github.com/yargs/yargs/pull/741#issuecomment-269588507

maybe we could make the CHANGELOG generation logic slightly different for a prerelease; perhaps it's as simple as:

1. default to adding a row to the CHANGELOG that does not enumerate bugs, patches, etc., and simply points out that the pre-release was released at a given tag:

-----

## 4.0.0-0 (2016-12-02) Prerelease

**pre-release 4.0.0-0**

-----

2. optionally allow people to specify a prerelease message during generation; apply the same GitHub ticket/username replacement when plopping this in the README.md.

CC: @Tapppi, @stevemao any ways! just brain dumping here 👍 not a blocker for this issue.

bcoe:
@fritzvd just swinging through some old issues, and prodding people, still think this is worth landing?

fritzvd:
hi @bcoe and @Tapppi I'm not sure you should land my PR. perhaps with the new changes @bcoe suggested for the pre-release it's kind of redundant?

Otherwise it might be better if @Tapppi refactors the content, because I don't think I could help out making it the way he should.
Or it would be so much of a back and forth, it would just be more efficient if @Tapppi changes it himself. Does that make sense?

bcoe:
@Tapppi any thoughts on this; maybe it would make sense to close this for the time being, and revisit once we address some of the new API changes in the works, e.g., lifecyle hooks.

bcoe:
this is a bit out of sync now, as we've been working on other features -- let's take another 🔪 at improving docs once we've landed a few of the neat features currently on the wire.

fritzvd:
👍🏼

PR #158, opened 2017-02-23
feat(options): add the ability to skip bump/changelog phase by using …

### LINKED ISSUE #157: How to bump/generate changelog and commit/push in a second time
Hello,

denouche:
…--skip-bump-and-changelog option, and to skip the commit/tag/push phase by using --skip-commit option

closes #157

### DIFF
--- a/command.js
+++ b/command.js
@@ -70,6 +70,18 @@ module.exports = require('yargs')
     default: defaults.tagPrefix,
     global: true
   })
+  .option('skip-commit', {
+    describe: 'Bypass the commit, tag and push to git phase. In another words just bump the version and create the changelog.',
+    type: 'boolean',
+    default: defaults.skipCommit,
+    global: true
+  })
+  .option('skip-bump-and-changelog', {
+    describe: 'Bypass the bump and changelog generation phase. In another words, just commit, create the tag, and push to git.',
+    type: 'boolean',
+    default: defaults.skipBumpAndChangelog,
+    global: true
+  })
   .version()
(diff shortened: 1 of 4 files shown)


Tapppi:
Lets wait for the discussion in #157 first, and then decide the fate of this!

stevemao:
I like the option, `skip-commit`, but not the other one :)

Tapppi:
@stevemao I agree, even if this is not necessary for htis use case once we get started on hooks, it might be worthwile anyway.

bcoe:
@Tapppi @stevemao @lancecaraccioli I like the idea of enumerating a set of lifecycles that occur in `standard-verison`:

`pre-commit`
`pre-changelog`
`pre-tag`

...

then it would be neat to allow any of these steps to be skipped within reason -- it would also be nice to try to isolate the logic for each so that, within reason, we can perform a rollback.

lancecaraccioli:
@bcoe I was thinking an incremental approach would make sense here.  Let's put the infrastructure in place to call hooks first. Then let the community add hooks  (via PR) based on their concrete use cases. I'll try to update my PR for #157 today based on discussion there.

bcoe:
@lancecaraccioli @denouche it's looking like the initial `hooks` work is close to landing; once this is polished with documentation, etc; I think we should then use this as the foundation for adding the support for skipping specific lifecycle hooks.

bcoe:
closing in favor of: https://github.com/conventional-changelog/standard-version/pull/188

PR #167, opened 2017-04-12
feat(lifecycle): add support for lifecycle hooks stor…

### LINKED ISSUE #157: How to bump/generate changelog and commit/push in a second time

lancecaraccioli:
Add hook support via "standard-version" package.json stanza

Closes #157

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,18 +32,24 @@ module.exports = function standardVersion (argv, done) {
     } else {
       checkpoint(args, 'skip version bump on first release', [], chalk.red(figures.cross))
     }
-
-    outputChangelog(args, function (err) {
+    runLifecycleHook(args, 'post-bump', newVersion, function (err) {
       if (err) {
+        printError(args, err.message)
         return done(err)
       }
-      commit(args, newVersion, function (err) {
+
+      outputChangelog(args, function (err) {
         if (err) {
           return done(err)
         }
-        return tag(newVersion, pkg.private, args, done)
+        commit(args, newVersion, function (err) {
+          if (err) {
+            return done(err)
+          }
+          return tag(newVersion, pkg.private, args, done)
+        })
       })
-    })
+    });
   })
 }
 
@@ -197,6 +203,15 @@ function handledExec (argv, cmd, errorCb, successCb) {
     }
     successCb()
   })
+}
+function runLifecycleHook(argv, hookName, newVersion, cb) {
+  var hookPath = path.resolve(process.cwd(), '.standard-version/hooks', hookName + '.js');
+  if (!fs.existsSync(hookPath)) {
+    cb();
+  }
+  handledExec(argv, 'node ' + hookPath + ' --new-version="' + newVersion + '"', cb, function () {
+    cb();
+  });
 }
 
 function commit (argv, newVersion, cb) {
--- a/index.js
+++ b/index.js
@@ -14,6 +14,7 @@ var objectAssign = require('object-assign')
 module.exports = function standardVersion (argv, done) {
   var pkgPath = path.resolve(process.cwd(), './package.json')
   var pkg = require(pkgPath)
+  var hooks = pkg['standard-version'] && pkg['standard-version']['hooks'] ? pkg['standard-version']['hooks'] : {}
   var defaults = require('./defaults')
   var args = objectAssign({}, defaults, argv)
 
@@ -32,7 +33,7 @@ module.exports = function standardVersion (argv, done) {
     } else {
       checkpoint(args, 'skip version bump on first release', [], chalk.red(figures.cross))
     }
-    runLifecycleHook(args, 'post-bump', newVersion, function (err) {
+    runLifecycleHook(args, 'post-bump', newVersion, hooks, function (err) {
       if (err) {
         printError(args, err.message)
         return done(err)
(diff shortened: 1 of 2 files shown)


Tapppi:
Hey, this is a great start! Sorry for the despicable turnaround time around `standard-version` lately, I think all the maintainers are really busy with projects and work right now. 

I think the hooks should live in a `package.json` config stanza, as discussed previously in https://github.com/conventional-changelog/standard-version/issues/157#issuecomment-282044903 . So the config would look like:

```json
{
  "standard-version": {
    "hooks": {
      "pre-commit": "node hook.js",
      "post-bump": "validate-artifacts"
    }
  }
}
```

@stevemao @bcoe What do you think about hook config, and what should the names for the hooks be?

lancecaraccioli:
Yeah I like that approach. I think it feels comfortable to follow git hooks lead on naming as most dev's would probably feel comfortable with that (keeping the learning curve really low)

bcoe:
@lancecaraccioli @Tapppi I like the idea of lifecycle scripts 👍 but I agree with @Tapppi that I think they'd make sense in a `hooks` stanza of `package.json`.

We use the yargs command line parser which supports most of this functionality already:

https://github.com/yargs/yargs#pkgconfkey-cwd

lancecaraccioli:
I'll be working on updating my PR today.

lancecaraccioli:
@Tapppi give this one another look please. Thanks

bcoe:
@lancecaraccioli this is looking great! sorry for the slow turnaround these days -- as @Tapppi has said himself, I think we're all buried under a few too many issues these days!

rather than the logic:

```js
var hooks = pkg['standard-version'] && pkg['standard-version']['hooks'] ? pkg['standard-version']['hooks'] : {}
```

I'd love to see us use the [pkgConf](https://github.com/yargs/yargs#pkgconfkey-cwd) feature of yargs -- this is what I use for nyc, and yargs itself. It provides some extra logic around finding the appropriate `package.json`, regardless of how the consuming library is installed -- this should pretty much be a drop in replacement for you 😄

lancecaraccioli:
Hey @bcoe I'll try to give this a look soon, hopefully this evening. Thanks

bcoe:
@lancecaraccioli if you have any cycles soon; I'd love to get some of these issues closed out.

index.js
```diff
@@ -32,16 +33,28 @@ module.exports = function standardVersion (argv, done) {
...
       }
-      commit(args, newVersion, function (err) {
+
+      outputChangelog(args, function (err) {
         if (err) {
           return done(err)
         }
-        return tag(newVersion, pkg.private, args, done)
+        runLifecycleHook(args, 'pre-commit', newVersion, hooks, function (err) {
+          if (err) {
```
SimeonC: Just a suggestion here, as this would help my own use case which I will outline below. If you allow the lifecycle hook to mutate/return a modified args struct that could allow more flexibility. (I'm currently considering using a `prepare-commit-message` hook to achieve this).

My company uses PivotalTracker which uses `[(Fixes|Finishes) #12312]` to indicate a story is fixed in a commit and then '[Delivers #12312]` to mark it as delivered and a github integration to detect the commit message and update the stories. What I'd want to do in this feature is change the `message` argument depending on all "fixed" commits since last version. (It'd be awesome if standard-version/conventional-changelog could pass all the commits it builds the changelog off to the hook as well but that might be asking for too much).

### Example
Git History:
```
fix: Some Issue [Fixes #123]
feat: Shiny Feature [Finishes #333]
chore(release): v1.1.2
```
On running `standard-version` I ideally want the release commit to be as follows so my stories automatically get updated to the delivered (for QA) state.
```
chore(release): v1.2.0

[Delivers #123]
[Delivers #333]
```
Assuming the hook allowed us to return a modified args I could append the new messages onto the `message` argument.

revin:
Hi all, I work on [marky-markdown](https://www.npmjs.com/package/marky-markdown); @bcoe recently helped get the project onto standard-version and suggested I throw my two cents in here.

This looks pretty great to me; in our case, for each release we build a small JSON file that includes the new package version number. We'd been doing that in the `version` npm run-script; I don't see a way to get standard-version to do that as is, but this PR looks like it would make that possible, so I'm into it 💯

In particular our use case is to have something run after the new version number is known, and to be able to programmatically generate changes to add to the commit that performs the version bump. So I'm guessing either of the `post-bump` or `pre-commit` hooks would work, yeah?

I don't have a strong or worthwhile opinion on decisions like how and where to specify the contents of the hooks themselves or anything; I'm just very interested in this feature overall 👍

bcoe:
@revin @SimeonC @lancecaraccioli please give this a shot:

`npm i standard-version@next` and see these docs:

https://github.com/conventional-changelog/standard-version#lifecycle-scripts

@SimeonC I tried to at least partially solve your use-case, you can provide a git message in the `precommit` hook -- perhaps you could look at the contents of CHANGELOG.md (which will have already been generated) to figure out the changes that have taken place?

bcoe:
@lancecaraccioli I've continued polishing the lifecycle script, documented here:

https://github.com/conventional-changelog/standard-version#lifecycle-scripts

Would love your help QAing, since the `next` release is fairly large, simply: 

`npm i standard-version@next`

PR #168, opened 2017-04-12
fix: add '-a' option to git commit command to actually commit all fil…

### LINKED ISSUE #166: document `commit-all`
Using the commitAll option does not actually commit all files.

Notice that the ternary based on `commitAll` results in an empty string instead of the expected `-a` commit option.

`(argv.commitAll ? '' : (argv.infile + toAdd))` 

The expected change

`(argv.commitAll ? ' -a ' : (argv.infile + toAdd))`

lancecaraccioli:
…es when commitAll option is true. Closes #166


lancecaraccioli:
closing as the solution to #172 will replace this.

PR #170, opened 2017-04-13
feat: add ability to specify conventional changelog preset

### LINKED ISSUE #169: Allow passing a an optional conventional-changelog "preset" option
Not all consumers of standard-version are on github, but would like to take advantage of the workflow provided by the standard-version project.  Allowing consumers to pass a conventional-changelog preset option will give them the ability to create custom configurations that suit their needs e.g. conventional-changelog-upwork

lancecaraccioli:
should close #169

### DIFF
--- a/index.js
+++ b/index.js
@@ -167,7 +167,7 @@ function outputChangelog (argv, cb) {
   }
   var content = ''
   var changelogStream = conventionalChangelog({
-    preset: 'angular'
+    preset: argv.conventionalChangelogPreset || 'angular'
   }, undefined, {merges: null})
     .on('error', function (err) {
       return cb(err)


bcoe:
@lancecaraccioli we're having a similar conversation on lerna:

https://github.com/lerna/lerna/issues/786

My concern is that tools like `conventional-recommended-bump` currently assume the `angular` commit style -- rather than introducing configuration toggles into standard-version. I'd rather keep perfecting this spec:

https://conventionalcommits.org/

@Tapppi thoughts?

lancecaraccioli:
My specific use case is that our repository is BitBucket and our issue tracker is Jira. The only other option would be to fork conventional-changlog-angular and maintain, what feels to be, a dirty copy.

by allowing conventional-changlog-VENDOR to be owned by VENDOR as an external repository, with the addition of this PR then it's possible to wire up to VENDOR specific resources.

bcoe:
@lancecaraccioli would love to support your use-case; so that I better understand -- does Jira have a specific commit format that it forces you to adhere to?

lancecaraccioli:
We have specific URL paths for user urls, bug links, diff links, etc.  I believe I made a minor change to a template. I'll past it here when I start work on polishing the docs for the PR

bcoe:
@lancecaraccioli okay cool -- what I was coming to; was maybe we could support this use-case by passing in an appropriate template context, and making some tweaks to the template 👍 (rather than adding more configuration toggles).

we could even publish a Jira preset 👓 using this new feature we've been building out in yargs:

https://github.com/yargs/yargs#extends-keyword

bcoe:
@lancecaraccioli I think we just need to add a couple optional parameters for specifying the issue URL, and to extend this functionality slightly:

https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-angular/index.js#L30

what do you think?

alan-agius4:
Duplicate of https://github.com/lerna/lerna/pull/1111

Jule-:
Definitely need this feature!
Indeed I was about to do exactly the same PR...

bcoe:
cleaning up some old pull requests, please feel free to reopen this conversation if you're still interested in implementing this.

PR #180, opened 2017-05-25
Release as exact version

stevemao:


command.js
```diff
@@ -7,7 +7,6 @@ module.exports = require('yargs')
     describe: 'Specify the release type manually (like npm version <major|minor|patch>)',
     requiresArg: true,
     string: true,
-    choices: ['major', 'minor', 'patch'],
```
stevemao: Ideally I like to keep this since it gives us really nice error messages 😸

bcoe:
@stevemao at some point in the last few days, it looks like there's been a regression specifically on Windows; I see this error on the mocha update too:

`handles commit messages longer than 80 characters`

I think one of us will need to spin up a VM and test things potentially.

bcoe:
@stevemao this feature is available in `standard-version@next`; I would love your help testing since `4.1.0` is a fairly large release.

`npm i standard-version@next`

PR #186, opened 2017-06-04
feat: adding support for lifecycle hooks

bcoe:
I'm putting the finishing touches on https://github.com/conventional-changelog/standard-version/pull/167

```md
`standard-version` supports lifecycle scripts. These allow you to execute your
own supplementary commands during the release. The following
hooks are available:

* `prebump`: executed before the version bump is calculated. If the `prebump`
  script returns a version #, it will be used rather than
  the version calculated by `standard-version`.
* `postbump`: executed after the version has been bumped and written to
  package.json. The flag `--new-version` is populated with the version that is
  being released.
* `precommit`: called after CHANGELOG.md and package.json have been updated,
  but before changes have been committed to git.

Simply add the following to your package.json, to enable lifecycle scripts:

```json
{
  "standard-version": {
    "scripts": {
      "prebump": "echo 9.9.9"
    }
  }
}
```
```

thanks for getting us most of the way there @lancecaraccioli.


lancecaraccioli:
Thanks for the polish @bcoe much appreciated. My work life interfered with my availability unfortunately glad to see you guys took it and ran with it.

PR #188, opened 2017-06-05
feat: add support for skipping lifecycle steps, polish lifecycle work

### LINKED ISSUE #157: How to bump/generate changelog and commit/push in a second time

bcoe:
## What is this?

this fairly major refactor addresses several feature requests around lifecycle hooks, and the ability to skip steps that standard-version performs.

There are now the events `bump`, `changelog`, `commit, and `tag`, each of which has `pre` and `post` lifecycle hooks, e.g., `prebump`, `postbump`, `precommit`, ...

fixes #157

## Note

I've branched from https://github.com/conventional-changelog/standard-version/pull/187 which should be landed first (so this commit looks bigger than it actually is).

CC: @lancecaraccioli, @denouche,

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,24 +156,37 @@ If you have your GPG key set up, add the `--sign` or `-s` flag to your `standard
 
 `standard-version` supports lifecycle scripts. These allow you to execute your
 own supplementary commands during the release. The following
-hooks are available:
+hooks are available and execute in the order documented:
 
-* `prebump`: executed before the version bump is calculated. If the `prebump`
+* `prebump`/`postbump`: executed before and after the version is bumped. If the `prebump`
   script returns a version #, it will be used rather than
   the version calculated by `standard-version`.
-* `postbump`: executed after the version has been bumped and written to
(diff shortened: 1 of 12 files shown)


bcoe:
I've pushed this to `next`, and have been testing on `marky-markdown` which has complicated build step:

https://github.com/npm/marky-markdown/blob/master/package.json#L14

I'm really happy!

```
✔ bumping version in package.json from 10.1.1 to 10.1.2
✔ Running lifecycle script "postbump"
ℹ - execute command: "npm run build"
✔ outputting changes to CHANGELOG.md
✔ Running lifecycle script "precommit"
ℹ - execute command: "git add marky.json"
✔ tagging release 10.1.2
```

denouche:
Nice!! Thank you!

denouche:
Maybe you could precise in the readme that the output of the `precommit` hooks is used as commit message. Because it's not obvious. By the way I'm not sure that this behavior is relevant, but it's you to decide. At least to have it in the documentation to avoid errors on releasing without understanding why.

Thank you!

Issue #146, opened 2016-11-26
pre-release improvements


bcoe:
The work done in https://github.com/conventional-changelog/standard-version/pull/129 by @e-cloud on pre-releases is awesome; I think things are 99% of the way there, and this will be a great feature.

I don't think the CHANGELOG generation is quite there yet though. 

## Current Behavior

Currently, if I release a pre-release, decide it's awesome, an then promote it to the actual release; it will result in a CHANGELOG that the entry `1.0.0-pre` with a description of all the changes, and `1.0.0` with no changes whatsoever, because the changes fall under `1.0.0-pre`.

## Ideal Behavior

Until the next actual release takes place, I think it would make sense to group all changes under an **Upcoming** header, we would then replace upcoming with whatever the next non-pre-release version is.

Is this the right behavior? other ideas? CC: @Tapppi, @e-cloud, @stevemao

Tapppi:
I think a side effect of the workflow/ideal behavior is that the changelog might include invalid information regarding a feature that has then been altered by a later feature. 

E.g. the changelog might include a `feat: add customName option` that later before a release is done is removed because it is entailed by `feat: add metadata option`. It'll be confusing to read about a new feature that actually doesn't exist anymore in the same form.

I can't come up with any easy way to solve this, other than allowing an option (`--preview`?) to edit the changelog before the commit of it..

bcoe:
@Tapppi put some thoughts here: https://github.com/conventional-changelog/standard-version/pull/152

stevemao:
Maybe we could duplicate some entries. EG:
A complete log from `v1.0.0` to `v2.0.0` would be before any `v2.0.0-pre` sections, and keep all `v2.0.0-pre` sections as is?`v2.0.0-pre`

stevemao:
Also wanna share something from http://keepachangelog.com/en/0.3.0/, and I followed the part "Unreleased" when I was working on conventional-changelog. The difference is that they don't use pre release but unreleased so this might be off topic...

[...]

PR #189, opened 2017-06-05
fix: Recommend --tag when publishing prerelease npm modules

### LINKED ISSUE #183: Recommend using --tag argument for prerelease in readme instructions
Nothing [this post from Mike Bostock](https://medium.com/@mbostock/prereleases-and-npm-e778fc5e2420), as well as the current behavior of npm, would it be preferable for standard-version to say

```
ℹ Run `git push --follow-tags origin master; npm publish --tag alpha` to publish
```

Instead of always

```
```

Which causes prerelease releases like 3.1.0-alpha.0 to be installed instead of stable releases?

tmcw:
The --tag argument to npm publish allows you to publish a prerelease without the prerelease assuming
the 'latest' tag and causing users to download the prerelease by default. Instead, we recommend
`--tag prerelease` by default when `--prerelease` is set in the `standard-version` call, so to
install a prerelease version of the module, users will install `module@foo` or specify the exact
version of the prerelease version.

Fixes #183

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,6 +21,7 @@ module.exports = function standardVersion (argv) {
   var newVersion = pkg.version
   var defaults = require('./defaults')
   var args = Object.assign({}, defaults, argv)
+  args[Symbol.for('messages')] = []
 
   return runLifecycleScript(args, 'prebump', null)
     .then((stdout) => {
@@ -50,6 +51,9 @@ module.exports = function standardVersion (argv) {
     })
     .then(() => {
       return tag(newVersion, pkg.private, args)
+    })
+    .then(() => {
+      return args[Symbol.for('messages')]
     })
     .catch((err) => {
       printError(args, err.message)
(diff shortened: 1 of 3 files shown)


tmcw:
Quick notes:

* Considered a few approaches for what to name the tag, but `prerelease` seems to make sense. standard-version --prerelease without an arg generates, for instance, 1.0.0-0, so simply tagging the semver tag (`npm install foo@0`) doesn't work that well. Similarly, `next` would imply a major version bump, I think, so `prerelease` is the closest to the desired meaning.
* There isn't test coverage for output. Adding it would be a more invasive change, so I don't include that in this PR.

stevemao:
Thanks a lot! Could you add a simple test for that?

tmcw:
Can do! I added the ability to test for output, by adding a Symbol key to the `args` object. It's the least invasive way and should avoid futurebugs around any new arguments, including arguments named `messages`. Open to other approaches if anyone sees other drawbacks to that testing approach.

bcoe:
@tmcw thanks for this 👍 unfortunately (fortunately) I just finished a [pretty big overhaul of `standard-version`](https://github.com/conventional-changelog/standard-version/blob/master/CHANGELOG.md#features) so I think you might need to rebase.

Would love your thoughts [on this old issue](https://github.com/conventional-changelog/standard-version/issues/146), regarding how we should generate a CHANGELOG for pre-releases.

stevemao:
@tmcw Is there a way to use `stdout` just like other tests?

tmcw:
Superseded by #196 which uses stdout to make testing much simpler.

Issue #135, opened 2016-10-28
Chores section


mrtnbroder:
When looking at https://github.com/conventional-changelog/standard-version/blob/master/CHANGELOG.md, I can see that there is indeed a Chores section, but it looks like I have to add an additional flag or something to generate it. Out of the box, I get no Chores section for my changelog generation.

Since the whole changelog generation is spread across thousands of repos (I'm exaggerating, but it feels much like it)
`conventional-changelog`, `conventional-changelog-core`, `conventional-changelog-cli`, `conventional-changelog-preset-angular` and what not, it's **really hard and frustrating** to understand and follow the codebase. Especially when trying to understand the README's. As every module is linking to some other module which actually has the information I want/need (maybe).

I could not find a flag or something that I could add to the  `standard-version` cli to generate the `Chores` section as well.

Tapppi:
Yes that chore is there because it is the cause of the Breaking Change. @stevemao @bcoe can you comment on this, I have no strong opinions on the subject, but I remember there being discussions about this.

nexdrew:
Typically, commits marked as **chore** do not end up in the CHANGELOG by default. Since the CHANGELOG file is [always included](https://docs.npmjs.com/files/package.json#files) in the published package, the idea is to strike a balance between including the important changes (in an easy to read/render format) and keeping the file to a reasonable size. If _all_ changes were included, it would essentially duplicate the entire git log.

As @Tapppi mentioned, the only reason that chore ended up in the CHANGELOG is because it was also marked as a **BREAKING CHANGE**, which is a little odd but understandable.

AFAIK, there is currently no way to tell `standard-version` to explicitly include chore commits (or anything other than **feat**, **fix**, and **BREAKING CHANGE**). This is the standard/convention, and `standard-version` aims to be an opinionated tool that encourages the use of that standard/convention. If you want the commits to be included in the CHANGELOG, you should use one of these labels.

That being said, I agree it would be nice to allow the convention to be overridden or augmented via CLI option, but I don't think `standard-version` is the tool to do that. (I also agree that the code is scattered amongst ETOOMANYPACKAGES, which _does_ make it difficult to add certain features to `standard-version`, but, for better or worse, this is the current state of the world until someone - like myself - has the time and energy to change it.)

stevemao:
see https://github.com/angular/angular/issues/5672
Normally chores don't make any difference to the end user of your app. You shouldn't make a release after only chores, style changes or code refactors.

[...]

PR #195, opened 2017-06-17
docs(FAQ): explain why is chore entries not in my CHANGELOG

stevemao:
I was surprised to see people asking for this since it made sense to me when I looked at it the first time. But too many people asking for it so maybe worth to add this? Mostly I copied @nexdrew's comment in issue https://github.com/conventional-changelog/standard-version/issues/135.

Closes https://github.com/conventional-changelog/standard-version/issues/135

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,14 @@ Although this will allow each commit to be included as separate entries in your 
 
 For this reason, we recommend keeping the scope of each PR to one general feature or fix. In practice, this allows you to use unstructured commit messages when committing each little change and then squash them into a single commit with a structured message (referencing the PR number) once they have been reviewed and accepted.
 
+### Why is `chore`, `style`, `refactor`, `docs`, `ci` or `build` entries not in my CHANGELOG?
+
+TL;DR: Those entries don't make any different to your end-users, unless there's a BREAKING CHANGE.
+
+Typically, commits marked as chore do not end up in the CHANGELOG by default. Since the CHANGELOG file is always included in the published package, the idea is to strike a balance between including the important changes (in an easy to read/render format) and keeping the file to a reasonable size. If all changes were included, it would essentially duplicate the entire git log.
+
+The only reason that chore ended up in the CHANGELOG is because it was also marked as a BREAKING CHANGE, which is a little odd but understandable.
+
 ## License
 
 ISC


README.md
```diff
@@ -307,6 +307,14 @@ Although this will allow each commit to be included as separate entries in your
 
 For this reason, we recommend keeping the scope of each PR to one general feature or fix. In practice, this allows you to use unstructured commit messages when committing each little change and then squash them into a single commit with a structured message (referencing the PR number) once they have been reviewed and accepted.
 
+### Why is `chore`, `style`, `refactor`, `docs`, `ci` or `build` entries not in my CHANGELOG?
+
+TL;DR: Those entries don't make any different to your end-users, unless there's a BREAKING CHANGE.
```
nexdrew: Maybe change this line to this?

```markdown
These entries shouldn't make any difference to your end-users, unless there's a BREAKING CHANGE, in which case the entry **will** be included in the CHANGELOG.
```

README.md
```diff
@@ -307,6 +307,14 @@ Although this will allow each commit to be included as separate entries in your
...
+
+Typically, commits marked as chore do not end up in the CHANGELOG by default. Since the CHANGELOG file is always included in the published package, the idea is to strike a balance between including the important changes (in an easy to read/render format) and keeping the file to a reasonable size. If all changes were included, it would essentially duplicate the entire git log.
```
nexdrew: Can we add this link?

```markdown
Since the CHANGELOG file is [always included](https://docs.npmjs.com/files/package.json#files) in the published package...
```

README.md
```diff
@@ -307,6 +307,14 @@ Although this will allow each commit to be included as separate entries in your
...
+
+The only reason that chore ended up in the CHANGELOG is because it was also marked as a BREAKING CHANGE, which is a little odd but understandable.
```
nexdrew: One more suggestion, let's change this line to:

```markdown
The only reason a chore (or similar entry) will end up in the CHANGELOG is if it is also marked as a BREAKING CHANGE.
```

stevemao:
Blocked by https://github.com/conventional-changelog/standard-version/issues/219

bcoe:
@stevemao this documentation seems reasonable, but my temptation is to potentially move it into a `docs/faq.md` file; Similarly, I was thinking we could have a `docs/advanced.md` outlining some interesting advanced tips and tricks.

We've taken this approach with yargs (splitting docs across a few files in `docs/`) and I think it's helped keep a more concise easy to follow README.md.

bcoe:
@stevemao let's dust off #219 and land this.

bcoe:
@stevemao @nexdrew do you still think this is worthwhile to add? any more comments @nexdrew.

hbetts:
I think it's still worth adding :+1:

bcoe:
@hutson @nexdrew  just bringing this ticket back to your attention, don't suppose we could reopen this conversation in slack?

hutson:
@bcoe I'm not in Slack anymore, but this PR still looks good to me. I feel the explanation will really help our community understand the versioning behavior of `standard-version`.

ruhkopf:
Hey everyone,

I was watching #280 which now refers to this ticket. Just to add my two cents to the discussion: We do have some smaller libraries, where we don't always add new features. Sometimes we just implement `doc` or `chore` changes. I'd argue many of those changes are still valuable for the end-user.

Take this very PR for example. If you would release this PR maybe along with some other documentation improvements, but no feature, then there would be a release without any release notes. This seems odd to the end-user.

Of course the workaround would be to just label such changes as a fix, but it would still be nice if I could configure the types of changes that go into the changelog based on my use case. I would assume this configuration would be different, depending on the type and size of the library/project/use case someone is working on.

Regarding difference to the git log: In our case, typically many commits make up a PR. The PR is then merged or squash merged with a conventional commit message. So this would be one entry in the change log per PR and not one entry per commit.

bcoe:
@ruhkopf @stevemao I'm seeing this same problem in some open-source libraries I contribute to at work; @stevemao how hard would it be to add a simple array configuration setting that allows someone to specify which sections are available in their README, with the default being the current behavior.

bcoe:
@ruhkopf want to create a ticket for this feature? I'd rather make this configurable than land these docs explaining why we don't support chores, to be honest.

stevemao:
@bcoe at the moment it is just a config file (not an option).

bcoe:
@stevemao what project does this config file live in; I think it would be nice to make it more easily configurable.

For context, @JustinBeckwith and I are exploring using conventional commit messages for some of the libraries we're working on. However, doing so will likely require a bit more flexibility in CHANGELOG generation, and some additions to the specification.

stevemao:
https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-core#config
you could `require` https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular, modify the config object and then pass it through `config` option.

stevemao:
There used to be a PR on adding a new preset to make angular-like conventions easy to config. I think we should create one and use it more.

bcoe:
@stevemao I agree, do you think we could make this the preset that we use for `standard-version` by default, and base it on: https://www.conventionalcommits.org/en/v1.0.0-beta.3/

How would this preset be different than others (what would make it more configurable)?

stevemao:
> I agree, do you think we could make this the preset that we use for standard-version by default, and base it on: https://www.conventionalcommits.org/en/v1.0.0-beta.3/

yes, and even angular config can be based on it

On top of my head, we can just list types that need to goto the CHANGELOG. BREAKING CHANGE keywords to mark this is important. Maybe some aliases (EG: `feat` becomes `Features` and `BREAKING CHANGE` will be listed under `BREAKING CHANGES` entry)? Pretty much https://github.com/Jimdo/changelogrc-spec

bcoe:
since https://github.com/conventional-changelog/standard-version/pull/323, you can now configure whatever commit types you'd like to show up in your CHANGELOG \o/

PR #196, opened 2017-06-23
fix: Recommend --tag prerelease for npm publish of prereleases

### LINKED ISSUE #183: Recommend using --tag argument for prerelease in readme instructions

tmcw:
The --tag argument to npm publish allows you to publish a prerelease without the prerelease assuming
the 'latest' tag and causing users to download the prerelease by default. Instead, we recommend
`--tag prerelease` by default when `--prerelease` is set in the `standard-version` call, so to
install a prerelease version of the module, users will install `module@foo` or specify the exact
version of the prerelease version.

Fixes #183 

Supersedes #189

### DIFF
--- a/lib/lifecycles/tag.js
+++ b/lib/lifecycles/tag.js
@@ -28,6 +28,7 @@ function execTag (newVersion, pkgPrivate, args) {
     .then(() => {
       var message = 'git push --follow-tags origin master'
       if (pkgPrivate !== true) message += '; npm publish'
+      if (args.prerelease !== undefined) message += ' --tag prerelease'
 
       checkpoint(args, 'Run `%s` to publish', [message], chalk.blue(figures.info))
     })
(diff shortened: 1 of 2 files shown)


tmcw:
Ping: anything else I need to do to get this merged?

spruce:
I would be interested to have it default to the name of the release one gives on the cli. Should I create a PR?

bcoe:
@spruce definitely, this sounds like a reasonable addition to this feature.

@tmcw this work is published to npm now, sorry that took so long!

PR #211, opened 2017-10-02
fix: use the `skip` default value for skip cli arg

spikegrobstein:
Even though the defaults for both `skip` and `scripts` were empty
objects, the code was referencing the default value for `scripts` in
both places, which could lead to a bug in the future if the default for
either changed at some point in the future.

### DIFF
--- a/command.js
+++ b/command.js
@@ -65,7 +65,7 @@ module.exports = require('yargs')
   })
   .option('skip', {
     describe: 'Map of steps in the release process that should be skipped',
-    default: defaults.scripts
+    default: defaults.skip
   })
   .option('dry-run', {
     type: 'boolean',


nexdrew:
@spikegrobstein Awesome, thank you for doing this! I too noticed this problem a while back but never bothered to fix it.

@bcoe @stevemao @Tapppi Any objections to merging this?

nexdrew:
@bcoe @stevemao @Tapppi Sorry to spam you, this is a simple fix, I will merge.

bcoe:
@nexdrew 👍 go ahead and publish.

bcoe:
@spikegrobstein this change is published now, sorry that took so long!

Issue #213, opened 2017-10-25
gitignored package-lock.json


satazor:
I prefer to gitignore the `package-lock.json` for libraries. I commit the `package-lock.json` for apps.

When the package lock is gitignored, when running `standard-version` this happens:

```
✔ committing package-lock.json and package.json and CHANGELOG.md
The following paths are ignored by one of your .gitignore files:
package-lock.json
Use -f if you really want to add them.

Command failed: git add package.json package-lock.json CHANGELOG.md
The following paths are ignored by one of your .gitignore files:
package-lock.json
Use -f if you really want to add them.
```

Using `-f` will add the package-log to source control which I don't want. Ideally, standard-version could detect that the file is gitignored and skip it.

Thoughts?

nexdrew:
@satazor Hey, I ran into this same problem today, bit of a pain.

Besides manually running `npm i --no-package-lock` (to avoid the creation of the lock file at all) or adding package-lock.json to the git repo, here are a couple workarounds/hacks I tested that were successful:

1. Rename package-lock.json on prebump and rename it back on posttag

    Define a standard-version lifecycle script in your package.json that looks like this:

    ```json
    "standard-version": {
      "scripts": {
        "prebump": "mv package-lock.json package-lock-IGNORE.json",
        "posttag": "mv package-lock-IGNORE.json package-lock.json"
      }
    }
    ```

    This maintains the integrity of the lock file but temporarily hides it when running standard-version.

    I put this as the first option because it still allows you to get the benefit of using package-lock.json file locally, though that value is arguably moot when not shared.

2. Remove package-lock.json on postshrinkwrap

    Define an npm postshrinkwrap script that looks like this:

    ```json
    "scripts": {
      "postshrinkwrap": "rm -f package-lock.json",
      "release": "standard-version"
    }
    ```

    This just subverts the normal lock file writing process that occurs with a typical `npm i` (npm will create the file and then immediately delete it on install). This works whether you're using standard-version or not.

    (Note that I also tried to do this with a postinstall script, but postinstall runs _before_ npm creates the lock file.)

3. Remove package-lock.json on prebump

    Define a standard-version lifecycle script in your package.json that looks like this:

    ```json
    "standard-version": {
      "scripts": {
        "prebump": "rm -f package-lock.json"
      }
    }
    ```

    This just removes the file when you go to cut a new release. The next time you run `npm i`, the lock file will be recreated.

[...]

PR #230, opened 2017-12-20
feat: do not update/commit files in .gitignore

bcoe:
I've bumped into this issue a few times, and noticed a similar conversation in #213.

How does this solution look? @nexdrew, @satazor, @arobson?


satazor:
Looks good, though tests are failing

lib/gitignore.js
```diff
@@ -0,0 +1,21 @@
+const Dotignore = require('@bcoe/dotignore')
```
nexdrew: For Node 4 support, this file needs `'use strict'` at the top. This fixes the failing tests for me locally.

nexdrew:
My attempt to fix the Windows build failed. Anyone else have ideas?

lib/gitignore.js
bcoe: @nexdrew I guess I'll need to spin up a Windows VM ... grumble.

bcoe:
@nexdrew I noticed the underlying library has some logic that uses `/` rather than `path.sep`; I might take @stevemao's advice and write our own little library that uses `find-up` and will patch the issue.

nexdrew:
@bcoe Are you talking about [this line](https://github.com/bcoe/dotignore/blob/90442b56563695be48bfced806faffc63e44ac4e/index.js#L29) (and maybe line 9 as well)? I don't really think this would affect the failing test, but we could try setting `matcher. delimiter = path.sep`.

nexdrew:
@bcoe I was able to spin up a Windows VM, and it looks like my suggestion of `matcher.delimiter = path.sep` actually works/fixes the failing test (which I find a bit surprising). Not sure my suggestion is the best long-term solution, but it might be enough to get this PR over the finish line.

lib/gitignore.js
bcoe: @nexdrew smart 👍

Aigeec:
Hey guys just wanted to let you know that I raised a bug on (dotgitignore)[https://github.com/bcoe/dotgitignore/issues/1]. It means that lines that are comments are included in the matches. So for example someone puts their directory name is as a comment the `dotgitignore` starts matching everything.

chrisgalvan:
@Aigeec Any update on when that will be fixed I just bump into that situation where my package.json and package-lock.json where ignored because some weird issue with https://github.com/bcoe/dotgitignore where it does match some ignored lines withe the full path of the package.json.

For example my package.json is at:
c:/user/myuser/dev/mypackage/package.json

and in my .gitignore I have the following
# ignore user files
.user*

because the path contains `user` package.json is being ignored.

Aigeec:
I've open a PR https://github.com/bcoe/dotgitignore/pull/3 on https://github.com/bcoe/dotgitignore to resolve the issue but I believe @bcoe is triaging it at the moment.

PR #232, opened 2018-02-06
doc : Added an example of a postcommit hook

spudheadc:
Added an example of a postcommit hook to replace links which refer to GitHub issues with a link to Jira issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,19 @@ Simply add the following to your package.json to configure lifecycle scripts:
 }
 ```
 
+As an example to change from using GitHub to track your items to using your projects Jira use a 
+`postchangelog` script to replace the url fragment containing 'https://github.com/`myproject`/issues/'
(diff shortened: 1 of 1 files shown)


stevemao:
You doesn't need to do this to achieve this. It's pretty hacky... But the config looks really simple.

spudheadc:
I agree it is not the best looking piece of config. And that a code solution would be better. However tthis particular hack means that my project didn't have to do a code change and could use standard-version out of the box without needing to fork. My requirements were simple - I wanted exactly what standard-version offers with the one exception, we use Jira rather than Github to track issues.

spudheadc:
@bcoe @stevemao in real life we are using $npm_package_bugs_url (see https://docs.npmjs.com/files/package.json#bugs) but this is linux specific and will not work in windows.  As I only run the standard-version in a linux based CI that will do for me. We could probably use something like https://www.npmjs.com/package/cross-var to do a cross platform implementation. 

Trying to document that however got tricky - so for the readme I stuck with the basics:-)

PR #234, opened 2018-02-22
feat: add prerelease lifecycle script hook (closes #217)

### LINKED ISSUE #217: prebump script with informational output causes version bump failure
I have an existing project I'm interested in using with standard-version. The catch is that it has a `preversion` script which runs JSDoc and commits updated API documentation before release; after going through the standard-version docs, I thought the `prebump` hook seemed analogous and made the appropriate changes. A dry run showed a proper changelog and the expected bump from 4.4.0 to 4.5.0. However, actually cutting a release did this instead:

```
✔ Running lifecycle script "prebump"
ℹ - execute command: "sh build/apidocs.sh"
✔ bumping version in package.json from 4.4.0 to null
✔ bumping version in package-lock.json from 4.4.0 to null
```

It then failed at the tagging step because `v4.4.0` already exists. Looking at [the code](https://github.com/conventional-changelog/standard-version/blob/master/lib/lifecycles/bump.js#L22) it appears that _any_ output in the `prebump` script is parsed as a version number, meaning that my original assumption was incorrect and `preversion` does _not_ map to `prebump`.

I think at minimum the docs could be clearer about this: the example `prebump` script echoes a version number, but the language "returns a version #" is a bit unclear in the context of a script and there's nothing indicating that _all_ of stdout is being parsed. I'm actually interested in my script's output so I don't want to pipe everything to /dev/null, and parsing just the last line in `prebump` seems feasible but kludgey; I think what's called for here is a separate lifecycle event (`prerelease` maybe?) which would be an actual analogue to `preversion` and run before `prebump`.

dmfay:
It took me a while to get around to it, but here's an implementation of a pre-everything script hook that won't affect behavior further into the process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,9 @@ If you have your GPG key set up, add the `--sign` or `-s` flag to your `standard
 own supplementary commands during the release. The following
 hooks are available and execute in the order documented:
 
+* `prerelease`: executed before anything happens. If the `prerelease` script returns a
+  non-zero exit code, versioning will be aborted, but it has no other effect on the
+  process.
 * `prebump`/`postbump`: executed before and after the version is bumped. If the `prebump`
   script returns a version #, it will be used rather than
   the version calculated by `standard-version`.
--- a/lib/lifecycles/bump.js
+++ b/lib/lifecycles/bump.js
@@ -20,7 +20,8 @@ function Bump (args, pkg) {
 
   if (args.skip.bump) return Promise.resolve()
   var newVersion = pkg.version
-  return runLifecycleScript(args, 'prebump')
+  return runLifecycleScript(args, 'prerelease')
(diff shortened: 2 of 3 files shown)


dmfay:
hey @bcoe -- any chance of this making it in sometime soon? :)

bcoe:
@dmfay I apologize for the slow turn around ... 2018 has been a bad year for OSS for me.

Refresh me, wouldn't attaching an event to `prebump` be equivalent to `prerelease`? What are the advantages of this; I suppose the fact that returning a version # doesn't have an effect on `releaseAs`?

dmfay:
No worries, these things happen. There's more information in the [original issue](https://github.com/conventional-changelog/standard-version/issues/217) but you've got the gist of it: any and all output from `prebump` is parsed for a version number, which makes it difficult to execute anything that produces its own output.

PR #236, opened 2018-03-04
feat: manifest.json support

fent:
Hi there,

I've added support for updating a `manifest.json` file, which is used in Google Chrome extensions. 
There's potential in standard-release to be used for software packages that are not node.js related or don't use npm. And I think this is a good start towards that.

Thank you for this tool btw, it's helped me a lot!

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,7 @@ const tag = require('./lib/lifecycles/tag')
 module.exports = function standardVersion (argv) {
   var pkg
   bump.pkgFiles.forEach((filename) => {
+    if (pkg) return
     var pkgPath = path.resolve(process.cwd(), filename)
     try {
       pkg = require(pkgPath)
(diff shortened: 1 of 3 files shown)


stevemao:
I guess we could merge this before we come up with a plugin strategy?

fent:
Just noticed I left the `.only` on the test. But now some other tests fail.

Will try to figure out why.

MAGALY956:
Y<a class="twitter-timeline" href="https://twitter.com/ThePSF?ref_src=twsrc%5Etfw">Tweets by ThePSF</a> <script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script>

bcoe:
@fent I like this refactor.

bcoe:
@fent please try, `npm i standard-version@next` it should have your changes.

PR #241, opened 2018-04-20
Show full tag name in checkpoint

MiniGod:
Tried this package out for the first time and wanted to undo what it had done.
When I tried to remove the tag, I assumed the output "tagging release 0.0.1" was the exact name of the tag, but it was `v0.0.1`.

This changes it so the output is representing the actual name of the tag, and not just the version.


bcoe:
@MiniGod please try `npm i standard-version@next` it should have your changes.

PR #243, opened 2018-05-14
Use tagPrefix in conventional-changelog

barmaglot92:
Use tagPrefix arg if passed also in conventional-changelog
If we won't to do it, the last commit won't find if we use custom tagPrefix


bcoe:
@barmaglot92 please try `npm i standard-version@next`.

barmaglot92:
@bcoe thanks a lot!

PR #275, opened 2018-11-01
feat: fallback to tags if no meta-information file found

bcoe:
_I'm working on some mad science with GitHub Actions, as part of this I'm working on making standard-version less tied to JavaScript._

This pull request makes it so that version will now be based on the git-history, if no supported meta-information file is found (package.json, composer.json, etc).

I've also updated the language slightly in the README to tie things less to npm.

BREAKING CHANGE: if no package.json, bower.json, etc., is found, we now fallback to git tags

CC: @TilmannBach

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,7 @@ module.exports = function standardVersion (argv) {
       if (!pkg && args.gitTagFallback) {
         return latestSemverTag()
       } else if (!pkg) {
-        return Promise.reject(new Error('no package file found'))
+        throw new Error('no package file found')
       } else {
         return pkg.version
       }
--- a/index.js
+++ b/index.js
@@ -1,3 +1,4 @@
+const latestSemverTag = require('./lib/latest-semver-tag')
 const path = require('path')
 const printError = require('./lib/print-error')
 
@@ -7,7 +8,7 @@ const commit = require('./lib/lifecycles/commit')
 const tag = require('./lib/lifecycles/tag')
 
 module.exports = function standardVersion (argv) {
-  var pkg
+  let pkg
   bump.pkgFiles.forEach((filename) => {
     if (pkg) return
     var pkgPath = path.resolve(process.cwd(), filename)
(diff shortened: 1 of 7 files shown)


lib/latest-semver-tag.js
```diff
@@ -0,0 +1,15 @@
+const gitSemverTags = require('git-semver-tags')
```
hbetts: Could this file be replaced by just using [`git-latest-semver-tag`](https://www.npmjs.com/package/git-latest-semver-tag)?

@stevemao could we also move `git-latest-semver-tag` into the `conventional-changelog` org? (since I seem to be stealing all your other packages recently :grin: )

test.js
```diff
@@ -2,16 +2,16 @@
...
+const shell = require('shelljs')
+const fs = require('fs')
+const path = require('path')
+const stream = require('stream')
+const mockGit = require('mock-git')
+const mockery = require('mockery')
+const semver = require('semver')
+const formatCommitMessage = require('./lib/format-commit-message')
+const cli = require('./command')
+const standardVersion = require('./index')
```
hbetts: These lines seem unrelated to this particular change. Could they be pulled out into a separate pull request that refactors the code base into modern ES6+?

index.js
```diff
@@ -7,24 +8,33 @@ const commit = require('./lib/lifecycles/commit')
...
+  let defaults = require('./defaults')
+  let args = Object.assign({}, defaults, argv)
 
   return Promise.resolve()
     .then(() => {
-      return bump(args, pkg)
+      if (!pkg && args.gitTagFallback) {
+        return latestSemverTag()
+      } else if (!pkg) {
+        return Promise.reject(new Error('no package file found'))
```
hbetts: Instead of returning a rejected promise, could you just throw the error here?

lib/latest-semver-tag.js
bcoe: I think I'd honestly rather add getting the latest tag as a feature to `git-semver-tags`, before we pull another repo into the mono-repo. It doesn't feel like there's enough of a functional difference between `git-semver-tags` and `git-latest-semver-tag` to warrant maintaining the additional library.

https://github.com/conventional-changelog/conventional-changelog/issues/388

test.js
bcoe: point taken re: not making more changes than necessary while adding a feature, at the same time since I pulled in a new dependency I usually tend to switch `var` to `const` for the surrounding imports (my feeling being it's better to eliminate this potential vector for bugs, rather than introduce a new vector).

I created a follow up ticket, to go through the codebase more thoroughly -- this feels like a great task for first time contributors:

https://github.com/conventional-changelog/standard-version/issues/276

Issue #43, opened 2016-05-23
Preserve formatting of package.json


nexdrew:
See discussion in #38, which was originally meant to address #37 but ended up in a slippery slope of scope creep.

Basically, we overwrite package.json using `JSON.stringify(pkg, null, 2) + '\n'` but this may be too intrusive for folks using Windows line endings (`\r\n` instead of `\n`) or different indentation (tabs instead of spaces, or more than 2 spaces).

Here are a couple ways we could address this:
1. Detect and preserve newline chars and indentation via [`detect-newline`](https://www.npmjs.com/package/detect-newline) and [`detect-indent`](https://www.npmjs.com/package/detect-indent) (similar to #38)
2. Read package.json as string (instead of `require()`ing it) and just replace the `"version"` value, ignoring/preserving all other content/formatting

I think I'd prefer option 2.

I also think our CHANGELOG writing logic should not be affected by this.

stevemao:
Probably should be similar to how `npm-version` does it.

bcoe:
@nexdrew might be worth looking at how greenkeeper does it.

nexdrew:
Looks like [`npm version`](https://github.com/npm/npm/blob/v3.9.6/lib/version.js#L283-L287) pretty much does the same thing that [`standard-version`](https://github.com/conventional-changelog/standard-version/blob/v2.3.0/index.js#L71) does. Hmm, maybe this is a non-issue then...

stevemao:
Is this still an issue?

nexdrew:
@stevemao I think we can close this. If anyone complains we can revisit.

ArmandoAssuncao:
@nexdrew I'm having problems with this, I use 4 spaces in the package.json, but it's changing to 2 spaces.
Could preserve the file, how you said.

tested in v4.3.0

stevemao:
I think we should reopen this since `npm` preserves the format now.

hugomrdias:
please preserve the code format, we are always fixing conflicts on package.json

rdbmax:
Can I submit a pull request for this ? Which option do you prefer ?
> Read package.json as string (instead of require()ing it) and just replace the "version" value, ignoring/preserving all other content/formatting

I am agree with this

alasdairhurst:
I've been looking for this for a while, so just got around to creating a PR for this: #282

bcoe:
@alasdairhurst @nexdrew is this still something anyone would like to take on?

[...]

PR #282, opened 2018-11-29
feat: preserve formatting when writing to package.json

alasdairhurst:
This PR implements a feature (or maybe it's a fix?) to get `standard-version` to preserve the indentation and 
line feed of `package.json` instead of overriding the user's preferred format with 2 space indentation. (See #43)

`detect-indent` and `detect-newline` are used to figure out the current indentation of `package.json`. This is done in an identical way to how `npm` handles writing changes to `package.json`.

Notable internal change:
1. `package.json` is no longer loaded with `require()`. While it should be functionally identical, in order to detect the formatting, the file to be read in as plain text. This caused certain tests to fail since the require cache wasn't updated during the bump, resulting in additional places where the cache needed to be deleted between invocations of the CLI in these tests. By using `fs.readFileSync()` in both places in the code where `package.json` is read instead of `require()`, the deletion of the `require` cache in the tests is no longer necessary for them to work.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,10 +1,10 @@
-const latestSemverTag = require('./lib/latest-semver-tag')
-const path = require('path')
-const printError = require('./lib/print-error')
-
 const bump = require('./lib/lifecycles/bump')
 const changelog = require('./lib/lifecycles/changelog')
 const commit = require('./lib/lifecycles/commit')
+const fs = require('fs')
+const latestSemverTag = require('./lib/latest-semver-tag')
+const path = require('path')
+const printError = require('./lib/print-error')
 const tag = require('./lib/lifecycles/tag')
 
 module.exports = function standardVersion (argv) {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -1,10 +1,10 @@
-const latestSemverTag = require('./lib/latest-semver-tag')
-const path = require('path')
-const printError = require('./lib/print-error')
-
 const bump = require('./lib/lifecycles/bump')
 const changelog = require('./lib/lifecycles/changelog')
 const commit = require('./lib/lifecycles/commit')
+const fs = require('fs')
```
alasdairhurst: Re-ordering requires to be alphabetical

test.js
```diff
@@ -95,16 +95,6 @@ function initInTempFolder () {
   shell.cd('tmp')
   shell.exec('git init')
   commit('root-commit')
-  ;['package.json',
```
alasdairhurst: No longer needed due to not using require() for metadata files

bcoe:
@alasdairhurst sorry this took so long to land, thank you for the contribution.

If you don't mind, give this a try:

`npm i standard-version@5.0.0`

You should find it has your changes.

PR #274, opened 2018-11-01
chore: update a few deps (fixing security issues)

bcoe:

PR #287, opened 2018-12-12
Bump deps

LinusU:
- Updates engines to `>=6` (missed doing that in #274 I guess?)
- Bumped dotgitignore which just removes Node.js 4 compat
- Removed fs-access which have been deprecated, it's in the standard fs package now

### DIFF
--- a/lib/lifecycles/changelog.js
+++ b/lib/lifecycles/changelog.js
@@ -1,4 +1,3 @@
-const accessSync = require('fs-access').sync
 const chalk = require('chalk')
 const checkpoint = require('../checkpoint')
 const conventionalChangelog = require('conventional-changelog')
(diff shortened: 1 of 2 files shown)


lib/lifecycles/changelog.js
```diff
@@ -52,7 +51,7 @@ function outputChangelog (args, newVersion) {
 
 function createIfMissing (args) {
   try {
-    accessSync(args.infile, fs.F_OK)
+    fs.accessSync(args.infile)
```
LinusU: `F_OK` is the default so removed that since `fs.F_OK` is also deprecated in favor of `fs.constants.F_OK`

bcoe:
@LinusU I apologize, I created a conflict with your branch while patching some security issues. Suppose I could get you to rebase?

bcoe:
renovate bot should now *hopefully* be setup to perform this action automatically.

Issue #294, opened 2019-01-16
if private module don't append `--tag prerelease`


whyvez:
`standard-version 4.4.0`

I've run into an issue with a prerelease for a private module where the `--tag prerelease` `npm publish` option was appended to the final `git push` help text.

```shell
✔ bumping version in package.json from 1.2.0-beta.4 to 1.2.0-dev.0
✔ bumping version in package-lock.json from 1.2.0-beta.4 to 1.2.0-dev.0
✔ outputting changes to CHANGELOG.md
✔ committing package-lock.json and package.json and CHANGELOG.md
✔ tagging release v1.2.0-dev.0
ℹ Run `git push --follow-tags origin master --tag prerelease` to publish
```

Need to wrap this logic within the if `(pkgPrivate !== true)` condition.

refs: https://github.com/conventional-changelog/standard-version/issues/183#ref-commit-709dae1

stevemao:
Thanks @whyvez , must be a bug.

PR #296, opened 2019-01-16
fix: no --tag prerelease for private module

whyvez:
fixes: #294

### DIFF
--- a/test.js
+++ b/test.js
@@ -440,6 +440,14 @@ describe('cli', function () {
       commit('feat: first commit')
       execCli('--prerelease alpha').stdout.should.include('--tag alpha')
     })
+
+    it('does not advise use of --tag prerelease for private modules', function () {
+      writePackageJson('1.0.0', { private: true })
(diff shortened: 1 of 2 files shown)


test.js
```diff
@@ -440,6 +440,14 @@ describe('cli', function () {
       commit('feat: first commit')
       execCli('--prerelease alpha').stdout.should.include('--tag alpha')
     })
+
+    it('does not advises use of --tag prerelease for private modules', function () {
```
stevemao: Still need to fix this: "does not advise"

stevemao:
Otherwise LGTM :)

test.js
whyvez: I had updated this but forgot to add it in my amended commit. (strike 2) 😬 Just forced pushed.

bcoe:
@whyvez thank you for this contribution, I'll roll it out with the security update 👍

bcoe:
@whyvez thank you for the contribution! Please give this a try:

`npm i standard-version@5.0.0`

If you'd like to contribute more to this project over time, don't be shy about joining the Slack:

http://devtoolscommunity.herokuapp.com/

PR #301, opened 2019-02-04
Register missing dependencies

medikoo:
`dependencies` in package.json doesn't list some direct dependencies of this project (now it works only by chance, that deep level dependencies are installed top level)

(as a side note: such errors could be avoided by relying on [import/no-extraneous-dependencies](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md) eslint rule)

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "dotgitignore": "^1.0.3",
     "figures": "^2.0.0",
     "fs-access": "^1.0.0",
+    "git-semver-tags": "^2.0.2",
     "semver": "^5.2.0",
     "yargs": "^12.0.2"
   },


stevemao:
Standard-version is not using conventional-changelog-angular directly now. conventional-changelog-angular is a direct dep of conventional-changelog. Currently it's configured via `preset` option. But I'd like to change it in the future.

medikoo:
>Standard-version is not using conventional-changelog-angular directly now. conventional-changelog-angular is a direct dep of conventional-changelog

Currently (on `master`) `conventional-changelog-angular` is expected to be installed as a dependency of `standard-version` otherwise most of tests will fail with `Unable to load the "angular" preset package`.

You may easily confirm on that by moving `conventional-changelog-angular` from top level `node_modules` into `node_modules/conventional-changelog/node_modules`, or just by installing project with npm v2 (which didn't do flat installs), and running tests.

And it's actually not only tests, invoking binary will fail as well with same error.

stevemao:
It might be a bug in https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-preset-loader/index.js then. 

I would like to remove `preset` option and import/require the config and just simply pass it to conventional-changelog... preset as a string unnecessarily complicates how to load config.

medikoo:
Sorry for confusion, I checked that again, and indeed `conventional-changelog-angular` doesn't have to be a direct dependency (when error occured on my side, most likely it was also not installed for `conventional-changelog`).

Still `git-semver-tags` is definitely missing. I've rebased PR

bcoe:
good find, I'm going to ahead and land this; as we do require that dependency and it's good to make it explicit.

PR #307, opened 2019-02-25
fix: prevent duplicate headers from being added (#305)

### LINKED ISSUE #305: "# Change Log" title gets inserted on every release
Despite following the instructions in the README, I get multiple titles inserted in my `CHANGELOG.md` file:

```md
# Change Log

All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.

# [0.2.0](https://github.com/ffxsam/xx/compare/v0.1.0...v0.2.0) (2019-02-19)


### Features

* **pirates:** Added the pirates ([929c0af](https://github.com/ffxsam/xx/commit/929c0af))



# Change Log

All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.

# 0.1.0 (2019-02-18)
```

jthomerson:
Fixes #305

### DIFF
--- a/lib/lifecycles/changelog.js
+++ b/lib/lifecycles/changelog.js
@@ -5,6 +5,7 @@ const conventionalChangelog = require('conventional-changelog')
 const fs = require('fs')
 const runLifecycleScript = require('../run-lifecycle-script')
 const writeFile = require('../write-file')
+const START_OF_LAST_RELEASE_PATTERN = /(^##? (?!Change Log$)|<a name=)/m
 
 module.exports = function (args, newVersion) {
   if (args.skip.changelog) return Promise.resolve()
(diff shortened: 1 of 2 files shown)


lib/lifecycles/changelog.js
```diff
@@ -5,6 +5,7 @@ const conventionalChangelog = require('conventional-changelog')
 const fs = require('fs')
 const runLifecycleScript = require('../run-lifecycle-script')
 const writeFile = require('../write-file')
+const START_OF_LAST_RELEASE_PATTERN = /(^##? (?!Change Log$)|<a name=)/m
```
jthomerson: Note that I really don't know of a bullet-proof way to do this because the format of where the "old content" begins is really dependent on the templates being used. For example, Angular may use either one `#` or two `##` depending on what kind of version it is (patch, etc).

ffxsam:
This wasn't an issue in standard-version 4.x. I wonder what the code looks like there for making sure it doesn't keep writing the title over and over.

jthomerson:
@ffxsam I didn't use the package at that version, so I'm not familiar with the headings it made, but it appears from looking at the code [1] that the headings contained `<a name=`. The code [1] was using `oldContent.indexOf('<a name=')` to find where the old content began.

[1] https://github.com/conventional-changelog/standard-version/blob/v4.4.0/lib/lifecycles/changelog.js#L26-L28

ffxsam:
Wait, looks like this was fixed in 5.0.1. I just tested & confirmed.

jthomerson:
@ffxsam I also confirmed that the fix in 5.0.1 seems to be working. Of course, that change (b684c784c8986226889feb1524bc029b913c9571) did not include any tests. So, you may still want to consider utilizing at least a portion of my PR here to add some tests so that you don't have a regression.

Also, both fixes have the same problem of not actually knowing how to parse a CHANGELOG since how you detect where the "old content" begins depends entirely on the format that's being used.

bcoe:
@jthomerson @ffxsam I'd honestly like to make the heading configurable (so that folks aren't completely tied to the default one); in the process, perhaps we could also make the regex that finds the last regex configurable -- tldr; I'd love to get this fixed, but maybe we should refactor the logic for putting the header in place a tiny bit.

bcoe:
@jthomerson thank you for the commit, I'm going to use this for a starting point and am going to make the header configurable 👍

PR #310, opened 2019-03-06
feat: Create --zero-mode flag to prevent bumping to 1.0.0.

### LINKED ISSUE #308: Support for 0.x semver mode
I have some a couple components that are currently on version `0.x.x` (but never `0.0.x`).  I'd like to propose a flag to enable `0.x` mode for version bumps.  Right now I have to manually use `standard-version -r minor` for breaking changes, `standard-version -r patch` for non-breaking features.  Bug fix releases already do the right thing.

To be clear, I want something like `standard-version --version-zero`, to block bumping to `1.0.0`, shift the meaning of semver: `0.breaking.(feat|fix)`.  npm allows a dependency on `^0.1.0` to install `0.1.1` but not `0.2.0`, so `^0.1.0` should allow updating to any non-breaking release from `0.1.0`.

coreyfarrell:
Fixes #308

### DIFF
--- a/lib/lifecycles/bump.js
+++ b/lib/lifecycles/bump.js
@@ -141,7 +141,16 @@ function bumpVersion (releaseAs, args) {
         path: args.path
       }, function (err, release) {
         if (err) return reject(err)
-        else return resolve(release)
+        else if (args.zeroMode) {
+          switch (release.releaseType) {
+            case 'major':
+              return resolve({ releaseType: 'minor' })
+            case 'minor':
+              return resolve({ releaseType: 'patch' })
+            default:
+              return resolve(release)
+          }
+        } else return resolve(release)
       })
     }
   })
--- a/test.js
+++ b/test.js
@@ -440,6 +440,52 @@ describe('cli', function () {
     })
   })
 
+  describe('zero-mode', function () {
+    it('works with patch release', function () {
+      writePackageJson('0.1.0')
+      commit('feat: first commit')
+      shell.exec('git tag -a v0.1.0 -m "my awesome first release"')
+      commit('fix: patch release')
+
+      execCli('--zero-mode').code.should.equal(0)
+
+      JSON.parse(fs.readFileSync('package.json', 'utf-8')).version.should.equal('0.1.1')
+    })
+
+    it('works with feature release', function () {
+      writePackageJson('0.1.0')
+      commit('feat: first commit')
+      shell.exec('git tag -a v0.1.0 -m "my awesome first release"')
+      commit('feat: feature release')
+
+      execCli('--zero-mode').code.should.equal(0)
+
+      JSON.parse(fs.readFileSync('package.json', 'utf-8')).version.should.equal('0.1.1')
+    })
+
+    it('works with breaking release', function () {
+      writePackageJson('0.1.0')
+      commit('feat: first commit')
+      shell.exec('git tag -a v0.1.0 -m "my awesome first release"')
+      commit('feat: breaking release\n\nBREAKING CHANGE: this is a breaking change')
+
+      execCli('--zero-mode').code.should.equal(0)
+
+      JSON.parse(fs.readFileSync('package.json', 'utf-8')).version.should.equal('0.2.0')
+    })
+
+    it('throws if not 0.x', function () {
+      writePackageJson('1.0.0')
+      commit('feat: first commit')
+      shell.exec('git tag -a v1.0.0 -m "my awesome first release"')
+      commit('feat: patch release')
+
+      const result = execCli('--zero-mode')
+      result.code.should.equal(1)
+      result.stderr.should.match(/can only be used with 0\.x\.x releases/)
+    })
+  })
+
   describe('manual-release', function () {
     it('throws error when not specifying a release type', function () {
       writePackageJson('1.0.0')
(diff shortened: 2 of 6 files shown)


test.js
```diff
@@ -440,6 +440,55 @@ describe('cli', function () {
...
+      commit('feat: first commit')
+      shell.exec('git tag -a v0.1.0 -m "my awesome first release"')
+      commit('feat: feature release')
+
+      execCli('--zero-mode').code.should.equal(0)
+
+      JSON.parse(fs.readFileSync('package.json', 'utf-8')).version.should.equal('0.1.1')
+    })
+
+    if (process.platform !== 'win32') {
```
JaKXz: I think the 

```js
if (isWindows) {
  return this.skip();
}
```

pattern we use in istanbuljs is a bit cleaner. I think it's worth the extra devDep

coreyfarrell: Done

panva:
npm handles versions below `1.0.0` as such that it won't download above the specified minor with the `^` operator (go ahead and [test](https://semver.npmjs.com/))

Because of that it's my assumption the norm below `1.0.0` is

```
breaking -> minor
features -> minor
fixes -> patch
```

and this PR should reflect that.

coreyfarrell:
I disagree about features.  If you specify `^1.0.0` you get non-breaking feature releases.  If you specify `^0.1.0` you should also get non-breaking feature releases.  That is why this PR implements zero-mode as:
```
breaking -> minor
features -> patch
fixes -> patch
```

panva:
While I can certainly understand the angle you're coming from, the fact that the API is deemed unstable (aka 0.x) does not remove the [specified](https://semver.org/) increment.

> MINOR version when you add functionality in a backwards-compatible manner

Regardless of what version you're on.

coreyfarrell:
I'm looking at [item 4](https://semver.org/#spec-item-4):

> 4. Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

This indicates to me that 0.y.z do not necessarily follow strict semver rules.  Specs aside the goal of this PR is to "do the right thing" for npm dependents.  From https://github.com/npm/node-semver/#caret-ranges-123-025-004:
> Many authors treat a 0.x version as if the x were the major "breaking-change" indicator.
>
> Caret ranges are ideal when an author may make breaking changes between 0.2.4 and 0.3.0 releases, which is a common practice. However, it presumes that there will not be breaking changes between 0.2.4 and 0.2.5. It allows for changes that are presumed to be additive (but non-breaking), according to commonly observed practices.

If I have a component at 0.1.0 and make a feature release followed by a bug fix release, another module which depends on `^0.1.0` should get the bug fix (released as 0.1.2, not 0.2.1).

bcoe:
@coreyfarrell I need this functionality to, but I think the right place to implement it is here:

https://github.com/conventional-changelog/conventional-changelog/pull/421

☝️ this provides the ability to provide configuration to the conventional bump logic, such that we'd treat a bump prior to the `major` differently; I think this should be the default behavior.

lib/lifecycles/bump.js
```diff
@@ -141,7 +141,16 @@ function bumpVersion (releaseAs, args) {
         path: args.path
       }, function (err, release) {
         if (err) return reject(err)
-        else return resolve(release)
+        else if (args.zeroMode) {
+          switch (release.releaseType) {
```
stevemao: This is technically not correct :) v0 is a bit complicated.

coreyfarrell: Sorry I'm not clear what you are saying is wrong?

stevemao: EG: there could be a breaking change from 0.0.1 to 0.0.2

coreyfarrell: Ahh true.  I hadn't considered `0.0.x` at all, I always just start projects at `0.1.0` as the way npm handles `0.0.x` is annoying to me (the caret in `^0.0.1` is useless).

bcoe:
@coreyfarrell having dug into things a bit more, I think it might be easier to change the bumping logic in `standard-version` itself (at least for the time being).

Alternatively we could refactor https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-recommended-bump to accept context about the old version we're attempting to bump from; which do you think is a better approach?

either way, I'd like the default behavior for pre-1.x.x to be bumping the minor for BREAKING CHANGES rather than the major.

coreyfarrell:
I'm not familiar enough with the conventional-changelog ecosystem to say if the refactor you mention is better or not. Modifying `standard-version` is definitely easier (possibly biased by my limited knowledge of everything involved).

I can modify this patch to remove the `--zero-mode` argument and just follow the zero-mode rules if the current version is 0.x.x (breaking changes = minor, feature or fix = patch).  The intent is that `standard-version -r major` and `standard-version -r minor` will not be effected by this change, this 'zero-mode' behavior will only happen if you do not specify an `-r` argument.  So `standard-version -r major` will bump from 0.x.x to 1.0.0.

What about 0.0.x, should this be handled differently as @stevemao pointed out?  Technically if the current version is 0.0.1 then 0.0.2, 0.1.0 or 1.0.0 could be used for breaking changes.  As is my patch will cause a bump to 0.1.0 instead of 1.0.0.

bcoe:
@coreyfarrell I still think this might be better fixed in `conventional-changelog`, it's just a bit more of a refactor than I thought, and I need to think on it a bit.

PR #318, opened 2019-03-28
fix(commit): don't try to process or to add changelog if it's skipped

georgiee:
If you skip changelog generation standard-version fails to process the remaining checkpoints. This PR provides a fix.

## Fixes
+ Prevent `CHANGELOG.md` or any given `infile` from being added to git when skipped.
+ Prevent `CHANGELOG.md` or any given `infile` from being added to the output when skipped.

## Problem Description 
Given this config:
```
"standard-version": {
    "skip": {
      "changelog": false,
      "tag": true
    }
  },
```

Before
```
✔ bumping version in package.json from 1.0.0 to 1.1.0
fatal: pathspec 'CHANGELOG.md' did not match any files

Command failed: git add package.json CHANGELOG.md
fatal: pathspec 'CHANGELOG.md' did not match any files
```

After 
```
✔ bumping version in package.json from 1.0.0 to 1.1.0
✔ committing package.json
```

Could not find any tests so I made the changes and tried it locally with the desired result.

Hope this works, thanks for this great tool.
Regards

### DIFF
--- a/lib/lifecycles/commit.js
+++ b/lib/lifecycles/commit.js
@@ -19,15 +19,26 @@ module.exports = function (args, newVersion) {
 
 function execCommit (args, newVersion) {
   let msg = 'committing %s'
-  let paths = [args.infile]
+  let paths = []
   let verify = args.verify === false || args.n ? '--no-verify ' : ''
   let toAdd = ''
+
+  // only start with a pre-populated paths list when CHANGELOG processing is not skipped
+  if (!args.skip.changelog) {
+    paths = [args.infile]
+    toAdd += ' ' + args.infile
+  }
+
   // commit any of the config files that we've updated
   // the version # for.
   Object.keys(bump.getUpdatedConfigs()).forEach(function (p) {
-    msg += ' and %s'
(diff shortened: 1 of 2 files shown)


georgiee:
Of course there are tests and of course I break them including some wrong formats. I'm going to provide a test and fix the code style issues.

stevemao:
Should we auto create a CHANGELOG if it doesn't exist?

georgiee:
I run into this problem while I set up versioning where I didn't need an actual changelog in the beginning. The config gives me the possibility to skip the changelog processing — which implies not to create a changelog when there is none yet. People would be really surprised to find a checked in and empty changelog after their decision to skip it for whatever reasons.

bcoe:
@georgiee seems reasonable to me, mind rebasing? sorry for the slow feedback.

georgiee:
Done 👌  All tests are running, my new test works too and the CLI behaves exactly as expected now.

igorsechyn:
Any chance this change can be released? This currently forces us to have an empty CHANGELOG file.

PR #319, opened 2019-03-28
fix: stop suggesting npm publish if package.json was not updated

MiniGod:
`standard-version@5.0.0` introduced git tag fallback, but it is still suggesting `npm publish` even if `package.json` was not updated.

Added a check so it doesn't. Could later be generalized for bower and composer, etc...?

Did a small refactor in `bump.js` so that the check would be simpler.

### DIFF
--- a/lib/lifecycles/bump.js
+++ b/lib/lifecycles/bump.js
@@ -162,10 +162,8 @@ function bumpVersion (releaseAs, currentVersion, args) {
  */
 function updateConfigs (args, newVersion) {
   const dotgit = DotGitignore()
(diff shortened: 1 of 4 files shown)


bcoe:
@MiniGod agreed, great addition; mind rebasing?

MiniGod:
rebased, but tests failing bc of lint errors of some other code (probably caused by the timing of #320 & #321). I'll rebase again when that's fixed on master.

bcoe:
@MiniGod I apologize for the slow turn around, `master` should be fixed again, mind rebasing one final time?

MiniGod:
It's ready for you now.

Issue #276, opened 2018-11-03
chore: update var to let/const


bcoe:
It might be worthwhile doing a bit of work to go through the codebase and update `var`s to `let` and `const` -- which have lexical scoping rules that make it harder to shoot yourself in the foot.

stevemao:
eslint should be able to auto fix this.

bcoe:
this has been done 👍

PR #321, opened 2019-03-30
chore: Updates `var` usage to `let`

jbottigliero:
- No longer uses `standard` directly, instead uses the [recommmend manual configuration](https://github.com/standard/eslint-config-standard) to allow for the addition of the [`no-var`](https://eslint.org/docs/rules/no-var) eslint rule.
- Uses `eslint . --fix` after the configuration update to swap `var` usage to `let`

see #276

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,6 @@
+{
+  "extends": "standard",
+  "rules": {
+    "no-var": "error"
+  }
+}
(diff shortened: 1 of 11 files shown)


jbottigliero:
This does move away from using the `standard` package directly, but it more closely resembles the `eslint` setup in [conventional-changelog](https://github.com/conventional-changelog/conventional-changelog) (with the addition of the `no-var` rule).

It looks like work might be picking up soon to get the rule introduced to `standard`, [here](https://github.com/standard/standard/issues/633). But this configuration would allow use to move forward and have one less `--fix` diff should we revert back to using `standard` directly.

PR #323, opened 2019-04-10
feats: adds new configurable conventionalcommits preset

bcoe:
Int his PR we introduce the conventionalcommits preset, here's why this is cool:

* the conventionalcommits preset is based directly on https://github.com/conventional-commits/conventionalcommits.org/pull/141, allowing us to better handle the slight deviations from angular's convention, e.g., the recent introduction of a `!` character for breaking changes.
* the conventionalcommits preset is highly configurable based on the [specification described here](https://github.com/conventional-changelog/conventional-changelog-config-spec).
  * this will allow `standard-version` to better support options other than GitHub in a well documented, standardized way.

BREAKING CHANGE: switches default preset to conventionalcommits

----

This change is currently blocked by:

* ~https://github.com/conventional-changelog/conventional-changelog/issues/435, which is preventing us from deploying new versions of conventional-changelog.~
* https://github.com/conventional-changelog/conventional-changelog-config-spec/issues/7, which would expose a list of configuration options for standard-version and other tools, so that it's easier to maintain consistent configuration options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [6.0.4](https://github.com/conventional-changelog/standard-version/compare/v6.0.3...v6.0.4) (2019-05-05)
+
+
+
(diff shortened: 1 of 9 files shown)


bcoe:
@jbottigliero thanks for your amazing work defining https://github.com/conventional-changelog/conventional-changelog-config-spec, I'm really happy with how things turned out.

Try out `standard-version@6.x.x`, it's now using your spec.

PR #347, opened 2019-05-05
feat!: bump minor rather than major, if release is < 1.0.0

### LINKED ISSUE #308: Support for 0.x semver mode

bcoe:
fixes #308 

BREAKING CHANGE: we now bump the minor rather than major if version < 1.0.0

### DIFF
--- a/lib/lifecycles/bump.js
+++ b/lib/lifecycles/bump.js
@@ -28,7 +28,7 @@ function Bump (args, version) {
     .then(runLifecycleScript.bind(this, args, 'prebump'))
     .then((stdout) => {
       if (stdout && stdout.trim().length) args.releaseAs = stdout.trim()
-      return bumpVersion(args.releaseAs, args)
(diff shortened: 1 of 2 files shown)


coreyfarrell:
This is great.  Probably tag with Fixes #308.

PR #351, opened 2019-05-08
fix: adds support for `releaseCommitMessageFormat`

### LINKED ISSUE #350: releaseCommitMessageFormat Support
This is a bit of a cross-repo (the change will likely occur in [the `conventionalcommits` preset](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-conventionalcommits))  ticket.

`releaseCommitMessageFormat` is not currently supported by the `preset`/`standard-version`. I've linked a PR with a failing test. 

---

~It seems like the fix will be to convert `releaseCommitMessageFormat` the [`message` property](https://github.com/conventional-changelog/standard-version/blob/master/lib/lifecycles/commit.js#L41) in the preset loader, but going that route does seem to couple the preset and standard-version in a not-so-obvious way.~ Added to `standard-version` by aliasing `--message`.

jbottigliero:
- adds support for `releaseCommitMessageFormat` from the spec via `--message` alias
- adds test covering: `.versionrc`, new `--releaseCommitMessageFormat` flag, and current `--message` flag

resolves #350


bcoe:
@jbottigliero this is looking good to me, perhaps keep this open and we can include the fix?

jbottigliero:
@bcoe – yup! Grabbing this one now.

command.js
```diff
@@ -26,7 +26,7 @@ const yargs = require('yargs')
     default: defaults.infile
   })
   .option('message', {
-    alias: 'm',
+    alias: ['m', 'releaseCommitMessageFormat'],
```
bcoe: maybe add a note that we'll deprecate message in the next `major`, and warn if people use `--message` or `-m`?

jbottigliero:
See #359 for the warning...

bcoe:
@jbottigliero mind rebasing this, this looks great to me.

jbottigliero:
@bcoe – updated, should be good to go.

bcoe:
@jbottigliero thank you!

PR #357, opened 2019-05-13
fix: update config file name in command based on README.md

mrastiak:
Changed the filename based on this section:
https://github.com/conventional-changelog/standard-version#configuration


bcoe:
@mrastiak thank you, good catch.

PR #359, opened 2019-05-14
fix: adds warning for -m and --message

jbottigliero:
- updates code to use releaseCommitMessageFormat but rewrite/warn in a check.
- updates format-commit-message to expect the new substitution pattern.

===

```bash
joe:standard-version $ standard-version --dry-run --message="foo %s"
[standard-version]: --message (-m) will be removed in the next major release. Use --releaseCommitMessageFormat.
```


lib/format-commit-message.js
```diff
@@ -1,16 +1,4 @@
...
-   * we have to make sure the substitution defined by the `conventional-changelog-config-spec`
-   * is handled properly.
-   * @see https://github.com/conventional-changelog/conventional-changelog-config-spec/blob/master/versions/1.0.0/README.md#releasecommitmessageformat-string
-   */
-  const message = String(rawMsg).replace('{{currentTag}}', '%s')
-  const matchCount = (message.match(/%s/g) || []).length
-  const args = Array(1 + matchCount)
-  args[0] = message
-  args.fill(newVersion, 1, args.length)
-  return util.format.apply(util, args)
```
jbottigliero: It seemed like the use of `util.format` was to support the old printf-esque (`%s`) replacement format... not sure if it also accounted for some other OS quirks I might be overlooking.

jbottigliero:
@bcoe, I had to take a slightly different approach to get the proper behavior around a warning message (due to the default). I opened this as a PR to my previous in case we want to take a slightly less invasive approach... let me know what you think.

bcoe:
@jbottigliero this looks great to me, thank you :+1:

PR #364, opened 2019-05-22
feat: Adds support for `header` (--header) configuration based on the spec.

jbottigliero:
- Adds shim to ensure fallback to the _current_ default (not the value in the spec)
- Adds warning for `--changelogHeader` usage.

```
joe:standard-version $ standard-version --dry-run --changelogHeader="Hello, World"
[standard-version]: --changelogHeader will be removed in the next major release. Use --header.
```


jbottigliero:
Resolves #364

bcoe:
@jbottigliero is the default behavior the same headers as before? just want to make sure we don't break a bunch of CHANGELOGs.

bcoe:
(to be fair, we look for the start of a version `## 3.x.x`  to decide on where the last CHANGELOG entry was, so I don't think changing the header would be likely to break anyone).

jbottigliero:
This shouldn't change any of the behaviors, just the sourcing... I _did_ have to move the validation that was occurring in the `.check` into the actual command handler in order to allow backward compatibility and proper overriding of the flag(s). 

Let me know if you think there is a specific case missing in the suite to add additional coverage.

PR #369, opened 2019-05-23
feat: validate CLI arguments using yargs.strict

### LINKED ISSUE #367: Check if arguments are valid before running release
It would be nice if standard-version would point out arguments that don't exist instead of running the release anyway.

For example: if I run `npm run release -- --dryrun` there should be a hint like `--dryrun is not a valid argument. Check your spelling and try again.` 

Not realizing that the argument is being ignored and probably running a second release is both frustrating and time consuming and could easily be prevented by checking the given arguments against the existing options.

jbottigliero:
```
$ standard-version --dryrun
Unknown argument: dryrun

Use --help for available options.
```

resolves #367

### DIFF
--- a/command.js
+++ b/command.js
@@ -105,6 +105,8 @@ const yargs = require('yargs')
       return true
     }
   })
+  .strict()
+  .showHelpOnFail(false, 'Use --help for available options.')
   .alias('version', 'v')
   .alias('help', 'h')
   .example('$0', 'Update changelog and tag release')


jbottigliero:
The tests here will fail due to some lifecycle [tests providing a `--patch` flag](https://github.com/conventional-changelog/standard-version/blob/master/test.js#L333).

https://github.com/conventional-changelog/standard-version/blob/a7133cc0e5a1924793bdf0e4abdd0ad9c58dfc2d/test.js#L319-L336

I'm not entirely sure what this `--patch` flag is used for and the tests seem to be passing without the flag being included. @bcoe would happen to have any insight here?

bcoe:
@jbottigliero I'm not sure what that flag is doing there; probably it's from a prior version of the CLI, and I bet it can be dropped.

jbottigliero:
Dropped the `--patch` flag, but I seem to be running into some issues around `yargs.strict()` which seems like it might be related to yargs/yargs#1039.

```bash
$ node bin/cli.js --dry-run --no-verify
Unknown argument: verify

Use --help for available options.
```

Issue #373, opened 2019-06-05
Unable to pass in `skip` to CLI


godd9170:
Attempting any of the following in the cli
```sh
standard-version --skip="{\"changelog\":true}"
standard-version --skip='{"changelog":true}'
standard-version --skip "{\"changelog\":true}"
standard-version --skip '{"changelog":true}'
```

returns the error `skip must be an object`

bcoe:
@godd9170 oh, perhaps ignore my comments on your PR, the issue is that you should be passing `--skip` in the format:

`--skip.changelog --skip.foo`

You can also provide these values in a configuration file.

godd9170:
Ahhh that makes way more sense. Thanks @bcoe

PR #375, opened 2019-06-06
fix(cli): Ensure that the skip and scripts cli arguments are first pa…

godd9170:
…rsed before they are evaluated

Without first JSON parsing the arguments of --scripts and --skip the typeof check within the
command.js was always throwing an error. It seems like yargs doesn't perform the parse for you, so
I've added the step to parse (and check for valid json). Now I'm able to do the commands that #373
is highlighting.

#373

### DIFF
--- a/command.js
+++ b/command.js
@@ -97,13 +97,28 @@ const yargs = require('yargs')
     describe: 'Commit message guideline preset (default: angular)'
   })
   .check((argv) => {
-    if (typeof argv.scripts !== 'object' || Array.isArray(argv.scripts)) {
-      throw Error('scripts must be an object')
-    } else if (typeof argv.skip !== 'object' || Array.isArray(argv.skip)) {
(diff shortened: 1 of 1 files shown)


bcoe:
@godd9170 thank you for the contribution 👍 I'll take a closer look when I have some cycles for open-source.

Mind adding a test, and taking a look at the failure that's occurring?

bcoe:
@godd9170 see my comments on your issue 👍 I think there's just a misunderstanding in how yargs handles objects; let me know if there's any clarification we could provide in the configuration section of the `standard-version` docs.

IllicLanthresh:
it'd be good if there was some clarification on how args work, i've been trying to pass an object and an array and I'm not able to do it from the cli, what's the correct way of passing a skip argument, just as an example, how would one skip commiting and tagging?

PR #377, opened 2019-06-06
fix(cli): Display only one, correct default for --preset flag

pioluk:
Previously `--help` would print both "angular" and "conventionalcommits" as defaults for `--preset`, which was confusing.

`$ standard-version --help`
Before:
```
  --preset             Commit message guideline preset (default: angular)
                                                        [string] [default:"conventionalcommits"]
```

After:
```
  --preset             Commit message guideline preset  [string] [default: "conventionalcommits"]
```

### DIFF
--- a/command.js
+++ b/command.js
@@ -94,7 +94,7 @@ const yargs = require('yargs')
   .option('preset', {
     type: 'string',
     default: defaults.preset,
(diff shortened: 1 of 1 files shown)


bcoe:
@pioluk thanks for the contribution 👍

PR #380, opened 2019-06-09
docs(cli): note about yargs dot notation in lieu of package.json

godd9170:
To help simpletons such as myself understand how yargs work (see #373)

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Making these URLs match GitLab's format, rather than GitHub's.
 
 ## CLI Usage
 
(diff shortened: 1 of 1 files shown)


bcoe:
@godd9170 my temptation, rather than drawing attention to yargs' docs, which should be an implementation detail, would be to just explain how dot notation works in our documentation.

godd9170:
Hey sorry for the delay. How's that wording instead?

bcoe:
@godd9170 thank you!

PR #391, opened 2019-07-04
fix:  remove unnecessary \n characters

moeshin:
Previously equivalent to only removing the first time.


bcoe:
@moeshin would happily land this if we can get tests passing, at a glance it seems like it changes the CHANGELOG format a bit.

moeshin:
@bcoe I don't know how to pass the test. 
#17 I think it can only delete the last \n character that was generated for the first time. So I made this modify.
Now I feel that it is unnecessary.  haha

PR #395, opened 2019-07-10
docs: add link to configuration spec

steveoh:
I had no idea what to put in the `.versionrc` until i stumbled upon the spec. A sample for how to do it in the `package.json` would be a great resource also but I don't know how to do that from reading the docs.

It is still unclear to me if it is only cli options are available or if the options in link I provided also work.


bcoe:
@steveoh any options available in the configuration spec should also be available as command line options :+1: 

Thanks for adding this documentation.

Please let me know if there's anything else we can add to the docs to help clarify, and I'll happily land more patches; appreciate the contribution.

PR #412, opened 2019-07-24
deps: bumps conventional-changelog to 3.1.9

jbottigliero:
- This will bump the transitive dependencies found in conventional-changelog which should ensure we're keeping in-sync with the features and fixes being applied to presets.
- This *did* require a change to a test case's expected output (-2 new-lines)... I don't think this needs to be seen as breaking and my assumption is it was just caused by some tweaks to templates in the conventional-changelog-conventionalcommits preset.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   "homepage": "https://github.com/conventional-changelog/standard-version#readme",
   "dependencies": {
     "chalk": "2.4.2",
-    "conventional-changelog": "3.1.8",
+    "conventional-changelog": "3.1.9",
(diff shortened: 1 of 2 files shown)


jbottigliero:
While this bump will fix some of the issues associated to [milestone/1](https://github.com/conventional-changelog/standard-version/milestone/1), ideally we get another bump/publish in `conventional-changelog` before releasing here.

bcoe:
@jbottigliero thanks for pulling together that milestone, it's exciting to see work happening here.

PR #413, opened 2019-07-24
docs(README): Removes refrences to Angular convention.

### LINKED ISSUE #406: Help info contains conflicting default values for --preset option
The help output for 6.0.1 states:

```
  --preset             Commit message guideline preset (default: angular)
                                                        [string] [default: "conventionalcommits"]
```

I gather [the default was changed in 6.0.0 to conventionalcommits](https://github.com/conventional-changelog/standard-version/blob/master/CHANGELOG.md#breaking-changes), but it looks like the reference to angular was not removed.

Thanks for this tool—it's great, and I am really excited to use it since #208 was fixed, as I want to ensure my changelogs are as accessible (e.g. to screen-reader users) as possible. Previously I had written a wrapper for conventional-changelog that modified the heading structure (and did some other stuff that's now supported via the new configuration options), but am looking forward to moving to using standard-version).

**Update:** was just looking at the README and noticed [the README also mentions that standard-version is based on the Angular format by default](https://github.com/conventional-changelog/standard-version#how-is-standard-version-different-from-semantic-release). If this is a separate issue, I'd be happy to file it.

jbottigliero:
- I just removed the reference to the Angular format for now... I'm not entirely sure the comparison/evaluation to `semantic-release` is still correct, but that seems like something we can evaluate _later_. 

closes #406

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,13 +351,13 @@ Tell your users that you adhere to the Conventional Commits specification:
 
 ### How is `standard-version` different from `semantic-release`?
 
-[`semantic-release`](https://github.com/semantic-release/semantic-release) is a fully automated library/system for versioning, changelog generation, git tagging, and publishing to the npm registry.
+[`semantic-release`](https://github.com/semantic-release/semantic-release) is described as:
 
-`standard-version` is different because it handles the versioning, changelog generation, and git tagging for you **without** automatic pushing (to GitHub) or publishing (to an npm registry). Use of `standard-version` only affects your local git repo - it doesn't affect remote resources at all. After you run `standard-version`, you still have the ability to review things and correct mistakes if you want to.
+> semantic-release automates the whole package release workflow including: determining the next version number, generating the release notes and publishing the package.
 
-They are both based on the same foundation of structured commit messages, but `standard-version` is a good choice for folks who are not yet comfortable letting publishes go out automatically. In this way, you can view `standard-version` as an incremental step to adopting `semantic-release`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -355,7 +355,7 @@ Tell your users that you adhere to the Conventional Commits specification:
 
 `standard-version` is different because it handles the versioning, changelog generation, and git tagging for you **without** automatic pushing (to GitHub) or publishing (to an npm registry). Use of `standard-version` only affects your local git repo - it doesn't affect remote resources at all. After you run `standard-version`, you still have the ability to review things and correct mistakes if you want to.
 
-They are both based on the same foundation of structured commit messages (using [Angular format](https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md)), but `standard-version` is a good choice for folks who are not yet comfortable letting publishes go out automatically. In this way, you can view `standard-version` as an incremental step to adopting `semantic-release`.
+They are both based on the same foundation of structured commit messages, but `standard-version` is a good choice for folks who are not yet comfortable letting publishes go out automatically. In this way, you can view `standard-version` as an incremental step to adopting `semantic-release`.
```
bcoe: I don't know that someone necessarily wants to move towards fully automated releases, as an example here's what we're doing currently at Google:

https://github.com/googleapis/release-please/pull/211

We create candidate PRs which an engineer can eventually land, perhaps we could tweak this wording a bit talking about this as a spectrum of automation that someone might have:

* manual work.
* release PRs like release please.
* continuous releases.

jbottigliero: I've reworded this a bit, let me know what you think!

Taking a step back, it looks like the comparison to `semantic-release` was added as part of the FAQ ~3 years ago... thinking about the expanded support for `bumpFiles` and movement toward the config-spec, I wonder if the README as a whole needs some finessing and if we could work toward dropping this section altogether. I _do_ see value in having a section devoted to common implementations/patterns to help folks better understand how "we" fit into the ecosystem, but maybe we can do it without specific comparisons.

PR #421, opened 2019-07-30
chore(deps,conventional-changelog): Bumps conventional-changelog to 3.1.10

### LINKED ISSUE #410: Links in changelog don't include repository name
The links to my published commits currently don't include a repository name, causing the links to not work.

For ex: https://github.com///compare/v1.0.0...v1.0.1 instead of
https://github.com/repository/name/compare/v1.0.0...v1.0.1

I have changed the name of my repository before, which I suspect could be the issue. Is there a way to configure the changelog so that future release links won't be broken?

jbottigliero:
- This is expected to resolve #410 based on updates in the conventional-commits preset.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   "homepage": "https://github.com/conventional-changelog/standard-version#readme",
   "dependencies": {
     "chalk": "2.4.2",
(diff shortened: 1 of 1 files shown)


towry:
This doesn't fix that issue !

PR #378, opened 2019-06-08
 feat(configuration): .versionrc.js files are now supported

jbottigliero:
- Updates the configuration retrieval to support Javascript (`.js`) configurations.
- Javascript configurations MUST export a configuration object _or_ a function (returning a configuration object) as the default export.
- Adds details around `.versionrc.js` to README.

closes #371

PR #437, opened 2019-08-18
feat(config): add .versionrc.js support

shirotech:


shirotech:
Actually, I just realised there is already a pending PR https://github.com/conventional-changelog/standard-version/pull/378 However, this one is much for simpler, whichever is decided to be merged is fine.

jbottigliero:
Hey @shirotech – thanks for the PR. 

As you mentioned this functionality is also included in #378. Comparing the code, the major difference (aside from some abstraction) is the handling of the export from the `.versionrc.js` file... #378 has additional logic to allow for exporting of a function:

https://github.com/conventional-changelog/standard-version/blob/77053647b593ead659220627a69b4e5f9e93a6b3/lib/configuration.js#L17-L26

I've gone ahead and rebased the original PR with the intent of landing it in `master` for the next release!

PR #448, opened 2019-08-30
feat: add ability to supply a version to bump from

mynameisstephen:
In some instances we may want to supply our own version to bump from, i.e. supply from a git tag or some other arbitrary source.

### DIFF
--- a/command.js
+++ b/command.js
@@ -16,6 +16,11 @@ const yargs = require('yargs')
     describe: 'make a pre-release with optional option value to specify a tag id',
     string: true
   })
+  .option('use-current-version', {
+    describe: 'Specify the current version to bump',
+    requiresArg: true,
+    string: true
+  })
   .option('infile', {
     alias: 'i',
     describe: 'Read the CHANGELOG from this file',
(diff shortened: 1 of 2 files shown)


jbottigliero:
Hi @mynameisstephen – thanks for the contribution. Do you think you could add a test covering the use of this flag? 

I'm thinking it might make sense to name this `--release-from`, this will better reflect it's purpose as the input version of `--release-as`.

TimothyJones:
Hi there! Since standard-version is deprecated, I've forked it [here](https://github.com/absolute-version/commit-and-tag-version). I'd like to accept this change, but I agree with the reviewer about the name of the parameter. Also, would you be able to add tests and documentation?

If you're still keen to see this released, I'd love a PR against the fork.

mynameisstephen:
> Hi there! Since standard-version is deprecated, I've forked it [here](https://github.com/absolute-version/commit-and-tag-version). I'd like to accept this change, but I agree with the reviewer about the name of the parameter. Also, would you be able to add tests and documentation?
> 
> If you're still keen to see this released, I'd love a PR against the fork.

Considering this pr is over 2 years old, I'd say no. Feel free to update it yourself if you want it.

TimothyJones:
That’s fair. I appreciate you taking the time to respond anyway

PR #495, opened 2019-12-05
No need to bumb version in composer.lock and composer.json

### LINKED ISSUE #394: Versions in composer
Hello! I'm about PR #262.

After any changes in composer.json command `composer validate` fail with error:

> The lock file is not up to date with the latest changes in composer.json, it is recommended that you run `composer update`

If use `composer update --lock` hash in composer.lock updated and section `version` removed from composer.lock

Docs:
https://getcomposer.org/doc/03-cli.md#validate
https://getcomposer.org/doc/03-cli.md#update-u

> --lock: Only updates the lock file hash to suppress warning about the lock file being out of date.

I added this trick in standard-version config:
```
"scripts": {
  "postbump": "composer update --lock"
}
```

In total:
1. Not need update composer.lock
2. Need run `composer update --lock` after changes composer.json, but command not very fast

Maybe it's need add to FAQ?

pvgnd:
closes #394

[Composer docs](https://getcomposer.org/doc/04-schema.md#version) states that version in composer.json is

> Optional if the package repository can infer the version from somewhere, such as the VCS tag name in the VCS repository. In that case it is also recommended to omit it.

### DIFF
--- a/defaults.js
+++ b/defaults.js
@@ -32,14 +32,12 @@ defaults.header = '# Changelog\n\nAll notable changes to this project will be do
 defaults.packageFiles = [
   'package.json',
   'bower.json',
-  'manifest.json',
-  'composer.json'
+  'manifest.json'
 ]
 
 defaults.bumpFiles = defaults.packageFiles.concat([
   'package-lock.json',
-  'npm-shrinkwrap.json',
-  'composer.lock'
+  'npm-shrinkwrap.json'
 ])
 
 module.exports = defaults


jbottigliero:
Hi @pvgnd, thanks for your contribution – this will need to be updated based on the changes to bump and package files that just landed in `master`, but this looks good to me and should be ok to release with the next major.

pvgnd:
Actually, doesn't seem to be the right fix. The hash of the composer.lock depends on the composer.json content, so after bumping the version in composer.json, the composer.lock is not update with the new hash. So we need to run a postbump hook : composer update --lock

jbottigliero:
@pvgnd maybe it makes sense to remove composer files from the default configuration altogether and provide an example for a composer-based library instead? Looking back on #394 and now that folks are better able to write their own updaters and bump files it might be the best route...

tommywo:
@jbottigliero I agree. Also [composer docs](https://getcomposer.org/doc/04-schema.md#version) states that version in composer.json is 

> Optional if the package repository can infer the version from somewhere, such as the VCS tag name in the VCS repository. In that case it is also recommended to omit it.

So we should also include that in the docs.

pvgnd:
I just rebased on master

pvgnd:
@jbottigliero I just rebased the PR

PR #511, opened 2019-12-18
doc: bumpFiles key for file path is not valid

antkaz:
When I updated the package to version `7.1.0` and added to config 
```
"bumpFiles": [
  {
    "file": "path/to/file.jsom",
    "type": "json"
  }
],
```
I get error `Path must be a string. Received undefined`. But when I changed `file` key to `filename`, then script is work


tommywo:
@antkaz Thank you for your contribution. 👍

PR #524, opened 2020-01-13
Fix incorrect key name in config

jozefcipa:
I've spent quite some time trying to figure out why my custom config didn't work out and I wanted to give up on the project but then I decided to take a look under the hood and found out there is incorrect information in `README.md` docs. So this PR fixes the wrong key name in JSON config.


tommywo:
Thank you for your contribution 👍
There already was previous pull request that address that, so I've merged it instead.
If you find more things that should be edited/added, we would be more than happy to take a pull request.

PR #527, opened 2020-01-16
docs(configuration): Fix typo & required versions

### LINKED ISSUE #271: customized path to package.json
Is there a possibility of adding customized path to package.json?
Let's say I have the following folder structure
untitled

First package.json (/project A) has name, scripts, dependencies and **NOT** version.

Second package.json (/project A/src/lib) has name, version, publishConfig and peerDependencies.

On executing standard-version in project root folder, the version needs to be updated accordingly to **src/lib/package.json**.
Currently it is taking root folder package.json which doesn't have version and updating the package.json by adding **version** key with **null** value.

Can it be customized to take path for package.json? If path not provided, behave as existing

### LINKED ISSUE #504: bumpfiles example uses "file" not "filename"
In the [Can I use standard-version for additional metadata files, languages or version files?](https://github.com/conventional-changelog/standard-version#can-i-use-standard-version-for-additional-metadata-files-languages-or-version-files) section of the README the `.versionrc` example uses `file` instead of `filename`.

I installed standard-version with `npm i -g standard-version@next` and have version `7.1.0`.

When the `.versionrc` refers to `file`, it prints this and returns 1.
```
The "path" argument must be of type string. Received type undefined
```

roschaefer:
I've spend at least an hour to figure out why my configuration files were always ignored. After debugging the source of my `node_modules/standard-version/*` it turns out I had `7.0.1` installed which apparently does not have the feature of `bumpFiles` or `packageFiles`.

I don't understand why `7.0.1` and not `7.1.0` got installed. I used `yarn add --dev standard-version`.

close #504
close #271

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ _How It Works:_
 
 `standard-version` will then do the following:
 
-1. Retreive the current version of your repository by looking at `bumpFiles`[[1]](#bumpfiles-packagefiles-and-updaters), falling back to the last `git tag`.
+1. Retrieve the current version of your repository by looking at `bumpFiles`[[1]](#bumpfiles-packagefiles-and-updaters), falling back to the last `git tag`.
 2. `bump` the version in `bumpFiles`[[1]](#bumpfiles-packagefiles-and-updaters) based on your commits.
 4. Generates a `changelog` based on your commits (uses [conventional-changelog](https://github.com/conventional-changelog/conventional-changelog) under the hood).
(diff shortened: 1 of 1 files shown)


jbottigliero:
Hey – thanks for catching this, while the README is expected to match the current release it looks like that's not the case.  

At the moment I do not have the proper permissions to publish the `npm` package... I'll work on getting this sorted out. It looks like the `7.1.0` release was tagged here on GitHub but not published in the registry.  It _does_ look like you might be able to use `standard-version@next` to get the proper version in your lock/package files. 

Thanks again for the contribution.

PR #541, opened 2020-02-06
docs: `packageFiles` is used to get version

MiniGod:
The readme states that `bumpFiles` is used to retrieve the current version, but it's actually `packageFiles`

https://github.com/conventional-changelog/standard-version/blob/2f04ac8fc1c134a1981c23a093d4eece77d0bbb9/index.js#L41-L43

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ _How It Works:_
 
 `standard-version` will then do the following:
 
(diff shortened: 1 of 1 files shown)


jbottigliero:
Hey – thanks for catching this! We had a few people catch this documentation error – the changes have landed in the trunk. If you believe additional changes should be made, feel free to open a new pull request.

Issue #525, opened 2020-01-14
repository url is no correct


lili21:
```bash
$ standard-version --dry-run
✔ bumping version in package.json from 0.2.2 to 0.2.3
✔ outputting changes to CHANGELOG.md

---
### [0.2.3](https://git.llsapp.com///compare/v0.2.1...v0.2.3) (2020-01-14)


### Features

* umd ([90fadf3](https://git.llsapp.com///commit/90fadf3d012a8427af65fb1582a5b7983956d2ed))
* umd ([694c577](https://git.llsapp.com///commit/694c577f9bb55277df7ee57a581756f7290abaf9))
* umd ([18644ac](https://git.llsapp.com///commit/18644ac313e935e156b67236034a364bc49c5e29))
* umd ([2dc3517](https://git.llsapp.com///commit/2dc3517bd86efcb89d4a0d1c52eebb08ba271745))


### Bug Fixes

* using idlize module and compile it ([a70c8ba](https://git.llsapp.com///commit/a70c8ba3d49f5c0298954e08cc56eaf27cb69bcd))
---

✔ committing package.json and CHANGELOG.md
✔ tagging release v0.2.3
ℹ Run `git push --follow-tags origin fix/idlize && npm publish` to publish
```

try to config it through `standard-verison` in package.json. not woking.

jbottigliero:
Hi @lili21! To better help troubleshoot your issue, could you provide:

- The output of `standard-version --version`
- Your `standard-version` configuration.

It'd also be worth verifying (and providing) the value of the `repository` property in your `package.json`. It's possible this is an issue with the preset _or_ the CHANGELOG generator...

lili21:
Here you go

```bash
$ standard-version --version
7.0.1
```

```json
// package.json
...
"repository": {
    "type": "git",
    "url": "https://git.llsapp.com/client-infra/thanos/thanos-web.git"
  },
...
```

lili21:
more information. it worked at first. but I transferred the repo, so I updated the git URL. After that, it never works.

aitazhixin:
i get the same error.

stevemao:
package url is handled in https://github.com/conventional-changelog/get-pkg-repo, which is just a wrapper of @hutson/parse-repository-url and hosted-git-info. you should probably look into them.

xtutu:
```bash
C:\Users\Administrator>standard-version -v
7.1.0
```

Same error when I use port in url.
```json
 "repository": {
[...]

Issue #227, opened 2017-11-29
feat: add configuration option for GitLab/BitBucket URLs


bcoe:
I would like to introduce config options to standard-version, in the form of a `.svrc` and a `standard-version` stanza in the package.json; This functionality is built into yargs:

https://github.com/yargs/yargs/blob/master/docs/advanced.md#configuration

I think the first feature we should support is adding the ability to use GitLab and BitBucket style URLs (see https://github.com/conventional-changelog/standard-version/issues/204, https://github.com/conventional-changelog/standard-version/pull/174).

@cdupetit could we potentially simplify the process of providing context for gitlab; it feels like `--context` requires a lot of variables be specified. I'd prefer that we override a few in the `package.json`.

bcoe:
CC: @Tapppi, @nexdrew, @stevemao

ahochsteger:
I was able to produce a working changelog for a self-hosted Bitbucket Server (v5.16.10) using the following configuration in .versionrc.json:
```json
{
  "types": [ ... ],
  "commitUrlFormat": "https://bitbucket.my.domain/projects/{{owner}}/repos/{{repository}}/commits/{{hash}}",
  "compareUrlFormat": "https://bitbucket.my.domain/projects/{{owner}}/repos/{{repository}}/pull-requests?create&targetBranch=refs%2Ftags%2F{{previousTag}}&sourceBranch=refs%2Ftags%2F{{currentTag}}",
  "issueUrlFormat": "https://jira.my.domain/browse/{{id}}",
  "userUrlFormat": "https://bitbucket.my.domain/users/{{user}}"
}
```
Note that since the comparison feature has been added with v6.3 of Bitbucket Server I'm generating links to create a pull requrest, which does not make much sense by itself, but at least there you can switch to the Diff tab where you can view the changes ;-).
You may wish to adjust the `compareUrlFormat` to use the real compare with v6.3 instead (where I don't know the URL format yet).

PR #174, opened 2017-05-10
feat(cli): Add a new context flag

cdupetit:
Running standard-version with the flag --context or -c allows you to specify a different host context.
```sh
# npm run script
npm run release -- --context <context.json>
# or global bin
standard-version --context <context.json>
```
In case you are using a private version control system installed locally like gitlab, and your repository url does not contain the git type keyword (github|bitbucket|gitlab). It is mandatory to inform manually standard-version on the host context (keywords) to be used.

package.json :
```json
{
  "name": "Private Project",
  "author": "Private Owner",
  "version": "1.0.0",
  "repository": "http:://private-repo.company.org/private-owner/private-project.git"
}
```
gitlab.json :
```json
{
  "issue": "issues",
  "commit": "commit",
  "referenceActions": [
   "close",
   "closes",
   "closed",
   "closing",
   "fix",
   "fixes",
   "fixed",
   "fixing"
 ],
 "issuePrefixes": [
   "#"
 ]
}
```

Issue #204, opened 2017-08-04
Changelog links are incorrect for selfhosted gitlab


AdamWillden:
I know the code attempts to best guess at the host type (gitlab, github, etc) based on the `repository` settings in the `package.json`, the changlog is then formatted with links as appropriate to the host type. Our gitlab server host name is very 'corporate' and unfriendly and does not contain the name 'gitlab'.

I suspect that [get-pkg-repo...index.js#L82](https://github.com/stevemao/get-pkg-repo/blob/ad3a6131b84fa5e85b1b5cbe4cfd0c4409c727a7/index.js#L82) is returning an `unknownHostedInfo` object. I don't know for sure.

I've looked through the code but cannot determine how best to fix this... I was hoping for something simple.

What would be great is if the user could specify what host type to use, and when not provided use the current auto-detection mechanism.

stevemao:
Is this the same as https://github.com/stevemao/get-pkg-repo/issues/9?

a-jackson:
Yes, I think it might be

AdamWillden:
@Stevemao i asked @a-jackson to look into it whilst I'm ok holiday. So what he says goes :-)

cdupetit:
@AdamWillden : Does this PR (#174) answer your question ?

AdamWillden:
@cdupetit yes that is it! #174 addresses this issue directly. 

Given we're both using gitlab I'd have been fine with a `--type gitlab` style fix but I'm absolutely fine with the `--context <context.json>` fix you have proposed there which gives much finer control and may prove useful.

I'd love to have this working!!

FWIW: I Inspected the changes in stevemao/get-pkg-repo#9 in the end myself and didn't see how it would fix our problem after all (I forgot to report that here however).

AdamWillden:
I'd be fine for you to reference this issue in that commit (such that merging will close this issue)

bcoe:
@AdamWillden please see comments in #227

bcoe:
@AdamWillden we finally have this functionality, as of `standard-version@6.x.x`, you can now create a `.versionrc` with custom URL formats [see](https://github.com/conventional-changelog/conventional-changelog-config-spec) for all the config settings that are available.

AdamWillden:
Thank you, very good of you to follow up and let me know.

PR #552, opened 2020-02-20
feat(changelog): allow to pass a custom context

dnlup:
In repositories that are self-hosted most of the time the changelog is genereated with broken links that are missing the owner and the repository in the path.
Allowing to pass a custom context to conventionalChangelog mitigates this.

See issue #525 and #227.

`.versionrc.js` example configuration
```js
module.exports = {
//...
context: {
  owner: 'some repo owner',
  repository: 'foo'
}
```

---

I saw #174 after opening this, sorry.

### DIFF
--- a/command.js
+++ b/command.js
@@ -4,6 +4,10 @@ const defaults = require('./defaults')
 
 const yargs = require('yargs')
   .usage('Usage: $0 [options]')
+  .option('context', {
+    type: 'string',
+    describe: 'Pass a custom context to inject during the changelog generation'
+  })
   .option('packageFiles', {
     default: defaults.packageFiles,
     array: true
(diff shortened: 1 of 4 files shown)


dnlup:
I am trying to figure out why the test fails on Windows. It will take a bit since I have to setup an environment for that. Any feedback is greatly appreciated in the meantime 🙏.

dnlup:
Maybe this work is not needed, I just saw https://github.com/conventional-changelog/standard-version/issues/204#issuecomment-489468725.

dnlup:
Since I opened this PR, I found alternative solutions for the problems I thought it would have solved. I still think there is value in adding this feature, but for now, I don't have the time to work on it. So I will close this to avoid unnecessary noise and come back later.

Issue #560, opened 2020-03-10
Feature request: file name in bumpFiles updater's script


Mualig:
Can we have the file name as a parameter for bump files updater's script?

I'm using templates and `sed` or `string.replace(...)` to update package version without having to write a parser for all the various file format I have, to find the version. The "bumped" filename would be useful to read content from the corresponding template.

PR #561, opened 2020-03-10
feat(bump): added filePath to writeVersion parameters

Mualig:
resolves conventional-changelog/standard-version#560


TimothyJones:
Hi there! Since standard-version is deprecated, I've forked it [here](https://github.com/absolute-version/commit-and-tag-version). I'd like to accept this change, but I think it would be better if the third argument to `writeVersion` were a metadata object with a key of `filePath` instead - this would allow better extensibility in the future without breaking changes or large function signatures.

If you're still keen to see this released, you're welcome to submit a PR against the commit-and-tag-version fork.

Issue #572, opened 2020-04-15
version sequence per branch


jahead:
Hey guys, 
 
With this now being able to be used for non-js projects. 
We were wondering if we could use this to manage our server environments. 

Use case
---

Master is a common template for a service 
Each branch is checked out from a master and one-to-one to a server environment where non-secret variables are set. 
E.G environment-specific variables in the settings.py in django.

The problem
---
Obviously the changelog problem is still a problem (but can fix that with `.gitattributes` or different CHANGELOG files).
We are mostly interested in having separate version lines for each environment
We currently are using --prefix-tag as we figured that could solve our version tag conflict problem, but we are getting an issue when we 'sync' (merge) from the master, the standard-version is getting confused even though `master => master/v1.0.1` and `prod => pro/v1.0.2`

it appears to ignore the --prefixTag when calculating the next version

If anyone has any extra information, or if your willing to let us patch this, please let us know. 

Thanks,
Joshua Head

PR #573, opened 2020-04-17
feat: allows seperate prefixTag version sequences

jahead:
Solves #572


jvitor83:
@bcoe need that PR merged and a new release please.

jahead:
there's conflicts I'll fix them on the weekend and update

jvitor83:
@jahead don't forget that please. :)

jahead:
Done. Sorry for taking so long. Let me know if anything else is needed

test/git.spec.js
```diff
@@ -66,7 +66,7 @@ function mock ({ bump, changelog, tags }) {
     }
   }))
 
-  mockery.registerMock('git-semver-tags', function (cb) {
+  mockery.registerMock('git-semver-tags', function (_, cb) {
```
bcoe, line 69: @jahead this is looking good to me, but any chance we could add a test that covers the additional functionality?

jahead: Yea mate, I"ll do it tonight. 

I was thinking how to test this

jahead: @bcoe  Hey so I added some tests here and tested tagPrefix in general, as there didn't seem to be any tests already that could see. 

I'm not overly happy with is this as ` git-semver-tags` actually is the one filtering the tags, and due to the mock we won't know if  `git-semver-tags` changes will break the logic here. But the tests cover the use case in #702 

resolve if you're happy

jahead:
Oddly, through testing this. 
I discovered/re-discovered that 
```
    if (pkg) {
      version = pkg.version
    } else if (args.gitTagFallback) {
      version = await latestSemverTag(args.tagPrefix)
    } else {
      throw new Error('no package file found')
    }
```

Which means that if there is a `packageFiles` defined we never get to use the gitSemverTags logic to resolve versions of different prefixes. And there's no option to force resolution by tag if `packageFiles` are defined. 

It will solve #702 but people should take care to configure no packageFiles. 

I'm not sure we should solve that problem in this PR, but it did seem like odd to me. 
I've outlined it in the tests.

bcoe (approved):

bcoe:
@jahead released in `v9.2.0`, and appreciated. Let me know if there are any issues.

I do gradually try to make my way through @ messages on GitHub, so feel free to keep engaging in this way.

Issue #590, opened 2020-05-12
Use version ranges in dependencies?


DullReferenceException:
I notice that the dependencies for `standard-version` are all fixed (no `^` or `~` for example). This makes it impossible to get the version bump in `conventional-changelog`, [which fixes a CVE](https://github.com/conventional-changelog/conventional-changelog/commit/46311d2932b367f370d06c4e447b8dcf4bc4e83f).

Could the `standard-version` dependencies be updated to use something like `^` so that upgrades and de-duplication of transitive dependencies is possible? If you object to this approach, could we at least get a new release of `standard-version` with `conventional-changelog` version bumped?

jbottigliero:
We've moved to allow semver ranges on a number of dependencies via #615 – we'll be working to unpin more as we phase our support of NodeJS@8 (#612, #618).

`8.0.1` was published ~6 hours ago which includes updates to `conventional-changelog` (#592).

PR #614, opened 2020-07-11
chore: Allow semver ranges for non-devDependencies – renovate-bot should no longer force the change.

jbottigliero:
Updates Renovate configuration to use "config:js-lib" (https://docs.renovatebot.com/presets-config/#configjs-lib), this introduces the ":pinOnlyDevDependencies" configuration.

see: #590

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "detect-indent": "6.0.0",
     "detect-newline": "3.1.0",
     "dotgitignore": "2.1.0",
-    "figures": "3.1.0",
+    "figures": "^3.1.0",
     "find-up": "4.1.0",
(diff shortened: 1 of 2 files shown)


bcoe:
@jbottigliero whoops, I addressed this independently [here](https://github.com/conventional-changelog/standard-version/pull/615). In an effort to give us some future proofing, I left some of our deps that have already dropped Node 8 pinned ... this is imperfect, open to better suggestions.

I think an approach of using `--strict-engine` in the future will help protect us from accidentally pulling in such hidden breaking updates.

jbottigliero:
Ah, that works and makes sense to me – I'll close this one out, thanks!

PR #616, opened 2020-07-12
test(windows): skip mock-git tests for Windows

bcoe:
The mock-git tests now exhibit a variety of odd behaviors on Windows:

* an error is captured and output regarding line endings.
* the commands ultimately passed to the `git` bin are in a different format (this is probably related to the switch to [`execFile`](https://github.com/conventional-changelog/standard-version/security/advisories/GHSA-7xcx-6wjh-7xp2).

tldr; let's turn off these tests on Windows for now.


bcoe:
cc:  @jbottigliero

PR #618, opened 2020-07-12
build: enable --engine-strict

bcoe:
This protects us from accidentally pulling in a version update that has an incompatible versions field.

Refs: #612


bcoe:
@jbottigliero we need to release the security patch first, then land #612, then we can land this.

jbottigliero:
With #612 landing and 9.0.0 released, I think we're good to go here.

PR #629, opened 2020-07-23
docs: changed bumpFiles to packageFiles for retrieval point 

rjmohammad:
The Readme defines `bumpFiles` as:

```
User-defined files where versions should be "bumped", but not explicitly read from.	- **`bumpFiles`** – User-defined files where versions should be "bumped", but not explicitly read from.
  - Examples: `package-lock.json`, `npm-shrinkwrap.json`
```

But the `Retrieval` point mentioned reading from`bumpFile` rather than `packageFiles`. 

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ _How It Works:_
 
 `standard-version` will then do the following:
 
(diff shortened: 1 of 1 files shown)


eemeli (approved):

jbottigliero (approved):
Thanks for your contribution!

PR #630, opened 2020-07-23
feat: support custom updater as object as well as path

### LINKED ISSUE #631: Using the same object in bumpFiles and packageFiles causes an error
**Describe the bug**
**Current behavior**

Using a config file like this:
```js
// .versionrc.js
const pom = {
  filename: 'pom.xml',
  updater: './path/to/custom-pom-updater'
}

module.exports = {
  bumpFiles: [pom],
  packageFiles: [pom]
}
```

Causes this error in `getCustomUpdater()`, because the `updater` field is overwritten by the first run of `resolveUpdaterObjectFromArgument()`:
```
The "path" argument must be of type string. Received an instance of Object
```

**Expected behavior**
Using the same object twice should not cause an error.

**Environment**
- `standard-version` version(s): v8.0.2, master
- Node/npm version: all
- OS: all

eemeli:
This adds support for defining the `updater` directly in a JS config file, rather than requiring it to be passed in as a path to be `require`d.

Fixes #631

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,7 +355,7 @@ As of version `7.1.0` you can configure multiple `bumpFiles` and `packageFiles`.
 
 1. Specify a custom `bumpFile` "`filename`", this is the path to the file you want to "bump"
 2. Specify the `bumpFile` "`updater`", this is _how_ the file will be bumped.
-  
+
     a. If your using a common type, you can use one of  `standard-version`'s built-in `updaters` by specifying a `type`.
 
     b. If your using an less-common version file, you can create your own `updater`.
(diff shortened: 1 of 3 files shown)


eemeli:
It looks like the force-push didn't trigger the GH action? That should have fixed the linter complaints.

eemeli:
Rebased on latest master, with test following new mock style.

brettdh:
I also ran into this issue today, in the context of expo-community/standard-version-expo#17.

In the [README.md in that diff](https://github.com/expo-community/standard-version-expo/pull/17/files#r487034017), I wrote `require.resolve` where I meant `require`, but the idea was to import the module and call it, returning a `{ readVersion, writeVersion }` object, as a way of passing extra parameters to the updater.

bcoe (approved):

brettdh:
Glad to see this merged. It's been a couple weeks since #662 was opened; when do you think we'll see 9.1.0 released? I'd like to start building on this functionality in expo-community/standard-version-expo#17.

PR #633, opened 2020-07-23
refactor: use async/await syntax to simplify source files

eemeli:
The current code style of this library is promise-based, using long chains of then/catch blocks. This PR replaces those in favour of async/await, which is well supported in Node.js 10, the minimum required by `standard-version`.

The logic of the code does not change here at all, but this allows for the code to have significantly less boilerplate syntax. Data flow is also simplified a little bit.


eemeli:
Rebased on latest master.

bcoe (approved):

PR #636, opened 2020-07-25
test: refactored test suite to use mocks

eemeli:
The test suite was really slow, to the extent that it made working on the library difficult. So I mostly fixed that here. Where previously it could take multiple minutes for it to complete, it's now at about 15 seconds total on my machine. And if you skip the commit & tag tests (now in `test/git.spec.js`), it's about 2 seconds. The coverage reporting does add a bit to that, of course.

Mostly the time savings come from not spawning wrapper processes around everything and mocking external dependencies and the filesystem whenever possible. A few tests are also dropped, as they were only testing external dependencies, rather than this library's own code.

The only non-test-code change is to `lib/updaters/index.js`, where the type-specific updaters are now loaded at the start, rather than on demand; this made mocking the filesystem around them significantly easier.

Code coverage does not change significantly during this change. Tests are split into three files, and moved to the `test/` directory.


eemeli:
Fixed the remaining Windows test issues.

Where previously the CI timings for the test suite took were 6min for the Ubuntu tests & a bit under 9min on Windows, it now takes 1.5 min & a bit under 4 min, respectively. I also added an `npm run test:unit` script, which skips the tests that use an actual git directory; that one completes in about 2 seconds.

bcoe (approved):
👍 this is a great improvement to the test suite.

lib/updaters/index.js
```diff
@@ -1,13 +1,17 @@
 const path = require('path')
 const JSON_BUMP_FILES = require('../../defaults').bumpFiles
+const updatersByType = {
```
bcoe, line 3: I like this refactor but it should probably be in a separate PR?

PR #638, opened 2020-07-27
feat: add cargo and cargo-lock updaters for Rust

viktor-ku:
Hi there!

I found myself using `standard-version` in CI for my Rust projects. I had to define updaters which are typical for any rust project, so I've decided to make a PR.

Now, `Cargo.lock` is not like `package-lock.json` in that, it does not have a fixed object with version, so I had to come up with the way to define a package name and reuse it when reading/writing the version. I hope that's fine.

Let me know if this can be landed in the near future, if at all. 

Thank you


alerque:
It's a bit disappointing to see that this work has been done and not merged and have zero feedback. What needs to happen in order to get this accepted? Tons of projects out here are re-inventing the wheel. If this isn't quite the approach you want to take, what is?

I just hacked up a simpler version of my own to get by until real support lands. In `package.json`:

```json
  "standard-version": {
    "bumpFiles": [
      {
        "filename": "Cargo.toml",
        "updater": "build-aux/cargo-updater.js"
      }
    ],
    "scripts": {
      "postbump": "cargo generate-lockfile --offline",
      "postcommit": "git add -u Cargo.lock && git commit --amend --no-edit"
    }
  }
```

And the aux script is:

```javascript
const TOML = require('@iarna/toml')

module.exports.readVersion = function (contents) {
  const data = TOML.parse(contents)
  return data.package.version
}

module.exports.writeVersion = function (contents, version) {
  const data = TOML.parse(contents)
  data.package.version = version
  return TOML.stringify(data)
}
```

PR #643, opened 2020-07-31
feat: support bumping python versions

yahehe:
Similar to https://github.com/conventional-changelog/standard-version/pull/638 and https://github.com/conventional-changelog/standard-version/pull/591, we've been adopting this tool in a predominantly Python codebase.

`setup.py`s aren't as structured as `package.json`, so just doing a regex find and replace. Could maybe be done smarter, but it works. `pyproject.toml` is a bit more structured but didn't want to add a TOML parser dependency, and this does the job.

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,6 +375,16 @@ As of version `7.1.0` you can configure multiple `bumpFiles` and `packageFiles`.
       "type": "json"
     },
     {
+      "filename": "setup.py",
+      // The `python` updater assumes the version is available under a `version` key in the provided setup.py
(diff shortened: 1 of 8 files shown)


Frank5000:
Any news on this?

djdembeck:
This would be useful to me as well.

theorange7:
Any news on this change being incorporated? Would love to have it support __init__.py files with __version__ = "v1.0.0" as well since that is also standard practice in some/most places 👍

TimothyJones:

PR #645, opened 2020-07-31
docs: updated generate artifacts code snippet

privatenumber:
### Updated the "Committing Generated Artifacts in the Release Commit" section
- **Removed issue reference** The issue is 4 years outdated and wasn't particularly helpful.
- **Corrected code snippet** The script didn't work (outdated?)
  - There is no hook called `release`
  - The script is `git add`ing the build files prior to the build that emits them. `git add` happens before `standard-version`, and `standard-version` is what calls the `prerelease` hook which builds the distribution files.


jbottigliero:
Thank you!

PR #649, opened 2020-08-05
feat: allow the tag message to appear as the title of releases in change log

gilesbradshaw:
if you set 
```
{
  "tagMessageInChangeLog": true
}
```

in .versionrc

and run standard commit with a tag message (or accept the default)

```
standard-version  --releaseCommitMessageFormat 'the final final version!'
```
you get the tag commit message for the release in the change log:

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,6 +7,7 @@ const path = require('path')
 const printError = require('./lib/print-error')
 const tag = require('./lib/lifecycles/tag')
 const { resolveUpdaterObjectFromArgument } = require('./lib/updaters')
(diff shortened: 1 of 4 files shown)


TimothyJones:
Hi there! Since standard-version is deprecated, I've forked it [here](https://github.com/absolute-version/commit-and-tag-version). I'd like to accept this change into the fork, but the PR includes a changelog for some reason, and it could do with some documentation for the new feature.

If you're still keen to see this released, would you be able to submit a PR against the commit-and-tag-version fork?

PR #666, opened 2020-10-02
fix: continue with commit when commitAll is true

wesleytodd:
When running with `commitAll` but also skipping `bump` and `changelog` steps, the commit is skipped.  This change gives what I assume is the expected behavior.


eemeli (approved):

TimothyJones:

PR #669, opened 2020-10-09
feat: provided lib lifecycle hooks run in context

### LINKED ISSUE #668: [Feature Request] Directly accept functions for life-cycle when used as library
When trying to use programmatically, it would be really nice to be able to extract data at the life-cycle hooks.

josh-hemphill:
closes #668
resolves #661
Went ahead and reused the old `run-lyfecycle-hook` file since it seemed fitting, and replaced its calls to place the new `run-lyfecycle-hook` as the middle-man and pass any available context to the provided lifecycle-hook functions.

As far as I can tell, this shouldn't be a breaking change since the datatype has to change in order to use it. If people think it would make sense to also have a config switch to turn it off, I'll add that.


josh-hemphill:
Never mind about the file, after tracking down the bugs I realized that all that functionality really belongs in one file.

jbottigliero:
Hey @josh-hemphill thanks for your contribution! I'm thinking rather than seeing this as a resolution to #661, we should drop `run-lifecycle-hook.js` (merging #661) and add the functionality required for #668 in _that_ handler,  potentially updating the name to simply `run-lifecycle` to avoid this feeling of having to go back-and-forth between the two. 

I know it's been awhile since your original proposal so please let me know if this is something you want to continue to work towards - thanks again!

josh-hemphill:
@jbottigliero 
> add the functionality required for #668 in _that_ handler

What's _that_ handler? It's imported by all four files under lifecycles.

jbottigliero:
Sorry, should have been a little more clear – 😅 !  

`lib/run-lifecycle-hook.js` was a deprecated utility that was no longer in use, it has now been removed from the trunk (see conflicts).

With that change in mind, it seems like we might not need to alter the call sites to achieve #668? Looking at the proposed code, the changes to `lib/run-lifecycle-hook.js` should actually be implemented in `lib/run-lifecycle-script.js`, with that change (and some test coverage), I think this would be good to go!

josh-hemphill:
The modification to the calls passes the version string to the hooks, which has been a lot more useful than the args.

Is there already tests specifically for library usage or that file?

jbottigliero:
This is looking good, a few things:

- I agree - the new and current version being passed to the lifecycle could be useful. I'm going to leave some inline comments on where there might be room for improvement on the proposed change.

- It looks like you might have merged in the release branch, not the trunk (`master`). You'll need to wait until after the 9.1.0 release lands and then update your branch again, or rebase against the trunk. We won't be able to merge this in as-is since it includes the version bump.

- **RE: Tests** You can find the existing lifecycle tests [here](https://github.com/conventional-changelog/standard-version/blob/master/test/core.spec.js#L228). I think just adding a single lifecycle test verifying the new arguments being passed when using a function would suffice – what do you think?

Thanks again for your work on this!

lib/run-lifecycle.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+module.exports = async function (ctx) {
+  const { args, hookName } = ctx
+  const scripts = args.scripts
+  if (!scripts || !scripts[hookName]) return Promise.resolve()
+  const command = scripts[hookName]
+  if (typeof command === 'function') {
+    checkpoint(args, 'Running lifecycle script "%s"', [hookName])
+    checkpoint(args, '- execute command: "%s"', [command.name], chalk.blue(figures.info))
+    return command(ctx)
```
jbottigliero, line 22: I really like the idea of passing the full context to the handler(s), I want to suggest a slight change to the shape of this parameter that I think will help structure the data being consumed by lifecycles and make it a little more obvious where/how to add additional properties on our end (without conflict):

```js
runLifecycle(name: string, { args: object, context: object })
// or
runLifecycle(payload = {name: string, args: object, context: object })
```

lib/lifecycles/tag.js
```diff
@@ -4,13 +4,13 @@ const checkpoint = require('../checkpoint')
...
-const runLifecycleScript = require('../run-lifecycle-script')
+const runLifecycle = require('../run-lifecycle')
 
 module.exports = async function (newVersion, pkgPrivate, args) {
   if (args.skip.tag) return
-  await runLifecycleScript(args, 'pretag')
+  await runLifecycle({ hookName: 'pretag', args: { ...args, pkgPrivate }, version: newVersion })
   await execTag(newVersion, pkgPrivate, args)
-  await runLifecycleScript(args, 'posttag')
+  await runLifecycle({ hookName: 'posttag', args: { ...args, pkgPrivate }, version: newVersion })
```
jbottigliero, line 13: With [this proposal](https://github.com/conventional-changelog/standard-version/pull/669/files#r528819190) in mind, I think `pkgPrivate` would actually belong to the payload `context`, not the `args` here since we derive it from a package file when available.

josh-hemphill:
I've got a bit more of a handle on how the tests are working now. I went ahead and added the API tests in their own section since it seemed like an anti-pattern to have the test hierarchy print out `CLI>lifecycle scripts>API`.

PR #670, opened 2020-10-11
fix: print version number that updater.writeVersion returns

brettdh:
For custom updaters that derive the new version from the file contents (e.g. `newVersion = oldVersion + 1`), the log output was showing the new computed semver version, rather than the actual version applied by the updater. This PR corrects the output for such custom updaters.

Closes expo-community/standard-version-expo#18.
Closes expo-community/standard-version-expo#10.


sreuter:
@jbottigliero Can we get this merged? It's causing some serious confusion at the moment.

TimothyJones:

Issue #242, opened 2018-05-11
Add switch to include commit body


derekgreer:
Currently standard-version only includes the commit header.  Is there a way to optionally include the commit body in the CHANGELOG.md?  Is there a sister tool I should be using to generate release notes to end users that provide all the details?

ffxsam:
@derekgreer Interesting no one noticed this, it's a great idea. When we squash our PRs on GitHub, it puts the squashed commits into the body, e.g.:

```
commit c18e83e1c45644ef7c81adae7fa7ed4581292b25
Author: Person
Date:   Wed Mar 20 16:46:13 2019 -0500

    feat(Downloads): Download modal improvements (#87)

    * fix: Some fix

    * fix: Updated things

    * fix: Did another thing
```
It would be cool if that body could be put into the changelog as an indented list underneath the main commit message.

ffxsam:
@bcoe @stevemao Thoughts on this?

willvincent:
Even just having extra detail beyond the first line, other than _only_ breaking changes would be useful.

Some new features, or involved fixes may warrant additional description above and beyond the one line.  Suggesting "you can manually edit the changelog" as has been mentioned in other issues, to my mind, kind of largely defeats the purpose of using an automated tool... if the info is _already in_ the commit, it should be able to be pulled out of the commit log and injected into the change log.

bcoe:
see motivation for a more flexible preset: https://github.com/conventional-changelog/conventional-changelog/pull/421

stevemao:
Maybe, we can use ‘<details>’ tag for commit body?

bcoe:
@stevemao this seems clever 😄 (in the good sense of the word). I might still be inclined to make it a config option that gets turned on.

joshtoper:
Was any progress made on this? This tool is fab but I require the full commit message in the Changelog, rather than just the header.

@stevemao Can you elaborate on what you meant re the `<details>` tag please?

stevemao:
@joshtoper see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details
I was just throwing random idea.

polaroidkidd:
I'm really in favour of this idea. I've done away with the develop branch in my projects as I tend to work myself from feature/issue to feature/issue and publish asap. Having the commit body included in the changelog would be very helpful in providing some context to the changelog.

Feiyang1:
[...]

PR #675, opened 2020-11-02
feat: add paremeter to consider own template files

mheunem:
This is a pull request for issue https://github.com/conventional-changelog/standard-version/issues/242


lib/lifecycles/changelog.js
```diff
@@ -32,11 +39,16 @@ function outputChangelog (args, newVersion) {
     }
     let content = ''
     const context = { version: newVersion }
+    const wirterOptions = {
```
leonyork, line 42: Should this be "writerOptions"?

mheunem: you are absolutely right.

leonyork: 👍 really looking forward to this being merged. Is there anything I can do to help?

PR #721, opened 2021-02-20
feat: allow updater's readVersion and writeVersion to execute asynchronously

nperez0111:
* Modifies signature of readVersion and writeVersion to allow Promise<string> as valid return

The reason for this change is that I have a use case where I need to fetch data asynchronously for putting into a file through an updater.
This should be backwards compatible without any modification to existing solutions. It simply makes a function execute asynchronously

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,13 +415,13 @@ module.exports = {
 
 An `updater` is expected to be a Javascript module with _atleast_ two methods exposed: `readVersion` and `writeVersion`.
 
-##### `readVersion(contents = string): string`
+##### `readVersion(contents = string): string | Promise<string>`
 
 This method is used to read the version from the provided file contents.
 
-The return value is expected to be a semantic version string.
+The return value is expected to resolve to a semantic version string.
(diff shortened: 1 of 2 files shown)


nperez0111:
Any progress on merging this in?

Akanoa:
I'd be really interested in this feature too :)

jirayupiwon1:
> * Modifies signature of readVersion and writeVersion to allow Promise as valid return
> 
> The reason for this change is that I have a use case where I need to fetch data asynchronously for putting into a file through an updater.
> This should be backwards compatible without any modification to existing solutions. It simply makes a function execute asynchronously

TimothyJones:
Hi there! Since standard-version is deprecated, I've forked it [here](https://github.com/absolute-version/commit-and-tag-version). I'd like to accept this change, but I think it needs tests with both the async and non-async versions.

If you're still keen to see this released, you're welcome to submit a PR against the commit-and-tag-version fork.

PR #748, opened 2021-04-15
adding YAML support

UnleashSpirit:
Hi,

As we working with some Helm charts, and using conventional commits, we also think that it could be a good idea to use standard-version on them but YAML support does not exist.
As Chart.yaml got a `version` key we decided to bring you this PR to support YAML file as general, assuming they have a `version` key like JSON updater

### DIFF
--- a/lib/updaters/types/yaml.js
+++ b/lib/updaters/types/yaml.js
@@ -1,0 +1,11 @@
+const YAML = require('yaml')
+
+module.exports.readVersion = function (contents) {
+  return YAML.parse(contents).version
+}
+
+module.exports.writeVersion = function (contents, version) {
+  const yaml = YAML.parse(contents)
+  yaml.version = version
+  return YAML.stringify(yaml)
+}
--- a/README.md
+++ b/README.md
@@ -322,6 +322,14 @@ npm run release -- --help
 standard-version --help
 ```
 
+### YAML support
+With CLI
+```sh
+# global bin
+standard-version --packageFiles path/to/your/file.yaml --bumpFiles path/to/your/file.yaml
+```
+Or using `.versionrc`, [see below](#can-i-use-standard-version-for-additional-metadata-files-languages-or-version-files)
+
 ## Code Usage
 
 ```js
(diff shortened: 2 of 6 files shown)


ferm10n:
Was also looking for a similar feature, but will likely just write an updater for standard-version to invoke.

What about the chart property `appVersion` too? `version` is just for the chart revision, but `appVersion` is for the actual version of the application to be deployed...

Maybe you could specify the property path to update instead of assuming it'll always be `version`, and have more control like this:
```json
    {
      "filename": "a/deep/package/dot/yaml/file/MY_VERSION_TRACKER.yaml",
      "type": "yaml",
      "propertyPath": "version"
    },
    {
      "filename": "a/deep/package/dot/yaml/file/MY_VERSION_TRACKER.yaml",
      "type": "yaml",
      "propertyPath": "appVersion"
    },
```

It does make me wonder though, whether this would need to handle when there's a prerelease like "2.0.0-alpha"... does `version` allow a string in helm? This is the default created by new helm charts:
```yaml
# This is the chart version. This version number should be incremented each time you make changes
# to the chart and its templates, including the app version.
# Versions are expected to follow Semantic Versioning (https://semver.org/)
version: 0.1.0

# This is the version number of the application being deployed. This version number should be
# incremented each time you make changes to the application. Versions are not expected to
# follow Semantic Versioning. They should reflect the version the application is using.
# It is recommended to use it with quotes.
appVersion: "1.0.0"
```

UnleashSpirit:
Hi, that could be a good idea
As you printed out, `Versions are expected to follow Semantic Versioning` so yes `2.0.0-alpha` as it's semver

**But I would LOVE that this PR (and this one ! #591)** to be merge before adding `propertyPath` as it's could be a generic property and not specific to yaml support

But it seems that this repo is dead ? Lots of issues/questions without any response, PR not followed (the java one is more than one year)
I'm so sad as I use it every day

Waiting for this to be merge, you can use my fork as it includes yaml and java support

lib/updaters/types/yaml.js
```diff
@@ -0,0 +1,11 @@
+const YAML = require('yaml')
+
+module.exports.readVersion = function (contents) {
+  return YAML.parse(contents).version
+}
+
+module.exports.writeVersion = function (contents, version) {
+  const yaml = YAML.parse(contents)
```
s7dhansh, line 8:
```js
const yaml = YAML.parseDocument(contents);
yaml.set('version', version);
return yaml.toString();
```

would be better as it would preserve the original comments, formatting etc.

TimothyJones:

PR #750, opened 2021-04-23
feat: detect if a release is necessary

### LINKED ISSUE #192: Detect if release is necessary, exit if not
`semantic-release` has a sweet way of detecting whether or not it should follow through with a release. Some commits are just docs, style, and even refactor, that don't really affect the published build, hence a release shouldn't really take place. Currently, `standard-version` tags the release regardless of what the commits are -- for commits that don't affect the build, `standard-version` bumps the version number as a patch. We use `standard-version` at work for our CI builds and it works great for publishing to our internal registry, but it overpopulates our registry with unnecessary releases. The same with Bitbucket and unnecessary release tags. 

Is this a viable feature that can be implemented into `standard-version`? Or is always releasing the correct behavior and should always be expected?

nelsonfncosta:
This PR tries to address the issue described in #192 in a way that is a non BREAKING CHANGE, the approach here is:

- Expose a new command option that enables this behavior `--detect-release`
- Get a list of commit since the latest git tag, via `git-log-parser`
- Check if there is any commit type worthy of a release(ex: `feat`), this is done by taking advantage of the configuration files(ex: `.versionrc`),  if no commit worthy is found we skip the release stage.

You can test this by doing a commit and then running the `release` command
ex:
```
...
git commit -m "docs: update README"
yarn release --detect-release
```

### Demo
https://user-images.githubusercontent.com/13886400/115864982-9713af80-a42f-11eb-8f8c-c373129a5db4.mov

### Notes
**Relevant links:**
https://github.com/bendrucker/git-log-parser
https://github.com/sindresorhus/get-stream

**Afterthought:**
As mentioned in the issue thread this behavior could eventually become a default, and a `--force` option could be added to give users the ability to always create a new release commit

Closes #192


nelsonfncosta:
Hey guys 👋 
I came across issue #192, and I tried to follow some ideas/suggestions in the thread.
This is somewhat a proof of concept, and also an opportunity for me to explore the codebase 😅 .
I'm ok with this being replaced/closed in favor of `--force` options as mentioned in the issue.

**Feedback is appreciated**  Thanks! 🙏

lib/utils.js
```diff
@@ -0,0 +1,15 @@
+const gitLogParser = require('git-log-parser')
+const toArray = require('stream-to-array')
```
satazor, line 2: Could we use https://www.npmjs.com/package/get-stream instead? It seems better maintained.

lib/utils.js
```diff
@@ -0,0 +1,15 @@
+const gitLogParser = require('git-log-parser')
```
satazor, line 1: Perhaps this file could be named `latest-commits`, which exports a `getLatestCommits` directly on `module.exports`? It seems other files are structured that way.

satazor (changes requested):

lib/utils.js
nelsonfncosta: Sure I can do that!
The reason I went with `stream-to-array` was because of the example/recommend way of doing it from `git-log-parser`

lib/utils.js
nelsonfncosta: The project seems to support both types of `exports` 
ex: 
https://github.com/conventional-changelog/standard-version/blob/3d341c24ba99920aafcefeed3a6bd9d156ea96fb/lib/configuration.js#L11

I'm gonna **hold** on this change for now, since my intentions with this `utils` file was to allow multiple named exports, if more utils should be added in the future people can still use the destruct assignment to get the desired function without having to refactor the default export 
ex: `const {util_1, util_2 } = require('./lib/utils')`

satazor: What I mean is that the author does have utils separated by files, instead of having a single utils file with all of them there.

index.js
```diff
@@ -33,6 +35,23 @@ module.exports = async function standardVersion (argv) {
...
+
+    const commitSubjectTypes = latestCommits.map(c => c.subject.split(' ')[0])
+
+    const releaseAvailable = commitSubjectTypes.some(c => releaseTypes.some(type => c.includes(type)))
+
+    if (!releaseAvailable) {
+      console.info(
+        chalk.green(`No commits found for types: [ ${releaseTypes.join(', ')} ], skipping release stage.`)
+      )
+      return 0
```
satazor, line 51: Why return 0 here? Also, I don't know if we should error out or exit successfully in this case. I'm leaned towards an error so that the process exists with != 0.

nelsonfncosta: The intention with the `return 0` was to say "script executed and exited successfully", I could replace it in favor of `process.exit()` 💭 

> Also, I don't know if we should error out or exit successfully in this case.

My perception of the issue at hand was that we didn't want to throw any error, what I understood from the initial issue was a use-case like _"I have CI that always runs `standard-version` when we a new commit is done, but it overpopulated some registry with unnecessary releases since some commits could be just `docs` or `styles` "_

I'm concerned that throwing an error could be something unnecessary alarming for the pipeline, but we can do that if other people agree.

lib/utils.js
nelsonfncosta: Got it! 
yes makes sense, I will refactor it

index.js
```diff
@@ -33,6 +35,23 @@ module.exports = async function standardVersion (argv) {
     }
   }
 
+  if (argv.detectRelease) {
```
bcoe, line 38: I would be tempted to pull this into a `detectRelease` helper method.

package.json
```diff
@@ -60,6 +60,8 @@
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^4.2.1",
+    "get-stream": "^6.0.1",
+    "git-log-parser": "^1.2.0",
```
bcoe, line 64: I think we could probably use `conventional-recommended-bump`, which is already pulled in as a dependency, and avoid pulling in new `get-stream` and `git-log-parser` deps.

test/utils.spec.js
```diff
@@ -0,0 +1,58 @@
...
+      expect(result).to.have.lengthOf(1)
+      expect(result[0].subject).to.equal('second-commit')
+    })
+  })
+
+  describe('when no tag is found', () => {
+    let result
+
+    beforeEach(async () => {
+      shell.exec('git commit --allow-empty -m "first-commit"')
```
bcoe, line 46: I appreciate the thorough tests 👍

bcoe (changes requested):
I like this feature, thank you for the contribution 👍 

My preference would be if we could figure out a way to `detect-release`, without pulling in new deps, I feel like we should be able to pull it off with some of the helpers already in the conventional-changelog monorepo.

package.json
nelsonfncosta: Quickly looking at the project, it seems to only expose a `releaseType`, do you think for a scenario where we only have commits with type `docs` , `conventional-recommended-bump` will return a `relaseType` of `null/undefined` opposed to `patch`?

Nonetheless, I will take a closer look into it!

index.js
nelsonfncosta: Just to confirm, are you suggesting to move the logic we have inside this `if` to it's own file and expose a `detectRelease` method that we could use call here?

package.json
nelsonfncosta: ok, I tested the following scenarios
having the following commit history  examples:
```
docs: update README
(tag: v9.2.0) chore: release 9.2.0 (#736)
```
running `conventional-recommended-bump -p angular` would result in `patch`

```
feat: added cool stuff
docs: update README
(tag: v9.2.0) chore: release 9.2.0 (#736)
```
running `conventional-recommended-bump -p angular` would result in `minor`

nelsonfncosta: Looking at options we might be able to take advantage of [whatBump](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-recommended-bump#whatbump)

What do you think of checking potential releases in the `whatBump` callback, the `commits` argument given to it seems to be equivalent information to what we get from `latest-commits.js`, so we would only need to check for release worthy commit and exit if none found (what we do with the `detectRelease` arg). 

https://github.com/conventional-changelog/standard-version/blob/280c4c0c073e220403814f94c79bbebb33a3c5f7/index.js#L47-L52

index.js
bcoe: @nelsonfncosta yes, I think it would make sense in its own helper method.

package.json
bcoe: @nelsonfncosta I like the idea of trying to use `whatBump`, we already do the parsing in the various parts of the conventional changelog repo, so I'm hesitant to pull in any other deps -- you might also be able to leverage some of the dependencies used within the conventional changelog monorepo already?

lib/checkpoint.js
```diff
@@ -3,7 +3,7 @@ const figures = require('figures')
 const util = require('util')
 
 module.exports = function (argv, msg, args, figure) {
-  const defaultFigure = args.dryRun ? chalk.yellow(figures.tick) : chalk.green(figures.tick)
+  const defaultFigure = argv.dryRun ? chalk.yellow(figures.tick) : chalk.green(figures.tick)
```
nelsonfncosta, line 6: Just wanted to highlight this change, `args.dryRun` seems to have been a typo.
Looking at the usage of `args` the expected is an array and not an object.

https://github.com/conventional-changelog/standard-version/blob/4383381f85b75cb35ca9e779f1c9b2505196aa10/lib/checkpoint.js#L8

package.json
nelsonfncosta: @bcoe, can you take another look when possible 🙏 

I was able to use `whatBump` and avoid pulling new dependencies, I split the new implementation into a separate commit to be easier to compare and see what changed.

bcoe (approved):

lib/lifecycles/bump.js
```diff
@@ -120,7 +131,34 @@ function bumpVersion (releaseAs, currentVersion, args) {
...
         preset: presetOptions,
         path: args.path,
-        tagPrefix: args.tagPrefix
+        tagPrefix: args.tagPrefix,
+        whatBump: (commits) => {
+          let level = PATCH
+          let breakings = 0
+          let features = 0
+
+          commits.forEach(commit => {
```
bcoe, line 140: I think this is most of the way there, but one concern, the conventional-commits strategy has a more complex `whatBump` implementation:

https://github.com/conventional-changelog/conventional-changelog/blob/8076d4666c2a3ea728b95bf1e4e78d4c7189b1dc/packages/conventional-changelog-conventionalcommits/conventional-recommended-bump.js

Which includes support for `!` in the commit body.

Is it possible for us to use whatever default `whatBump` is configured already, and then update the `reason` afterwards?

bcoe (changes requested):
Sorry for additional nit, this is looking most of the way there.

lib/lifecycles/bump.js
nelsonfncosta: Initially, I had that approach in mind but after some exploring, I was not able to find a clear way to get the default `whatBump` implementation, ending up implementing my custom `whatBump` :/ 

💭  Would be cool if `conventional-recommended-bump` passed the default `whatBump` callback  as an option argument  to the `whatBump` option, `whatBump(commits, { defaultWhatBump, ...}) {};` 💭 
[It send a bunch of options](https://github.com/conventional-changelog/conventional-changelog/blob/8076d4666c2a3ea728b95bf1e4e78d4c7189b1dc/packages/conventional-recommended-bump/index.js#L86) but none we can use.

How do you think we should proceed @bcoe ?
I can adapt the current implementation to follow the strategy you linked, to support the `!` use case you mentioned.

bcoe: @nelsonfncosta I think each preset should be exporting its `recommendedBump` configuration, see:

https://github.com/conventional-changelog/conventional-changelog/blob/8076d4666c2a3ea728b95bf1e4e78d4c7189b1dc/packages/conventional-changelog-conventionalcommits/index.js#L29

The problem with copying the logic over from any one preset, is that we allow folks to configure their own preset -- so although the logic would be good for conventional commits, it would be off potentially for another plugin.

I haven't tried to do this myself, and apologize for how much of a yarn ball the conventional changelog codebase is -- I think though you should be able to pluck `whatBump` off of a given preset.

nelsonfncosta: I will take a closer look into this! 👀 
The information we need is on the `recommendedBumpOpts`, so if this is being exported it should be a straightforward change 👍

nelsonfncosta: I have been looking for a way to get the preset config
something along the lines

```
const options = {} // mostly the info in args 
const config = require('conventional-changelog-conventionalcommits')

const { recommendedBumpOpts } = config(options)
```

the problem I'm facing I'm not really sure how to "wait/get" my value from [here](https://github.com/conventional-changelog/conventional-changelog/blob/8076d4666c2a3ea728b95bf1e4e78d4c7189b1dc/packages/conventional-changelog-conventionalcommits/index.js#L36)

Vyazovoy:
Would be really good to merge this PR as it fixes significant flaw.

patoncrispy:
Hey there, just wondering if there is much left in this PR? Is there help needed getting this across the line?

PR #759, opened 2021-05-19
expose --skip-unstable flag from conventional-recommended-bump

ferm10n:
conventional-recommended-bump supports a [`skipUnstable`](https://github.com/conventional-changelog/conventional-changelog/blob/3b0fb22b0bb49137b9ecddaeb349f66025205c57/packages/conventional-recommended-bump/README.md#skipunstable) option which would be very useful for projects that use standard-version and want to make release candidates

I wanted to have this feature supported so I could:
- make a v1.0.0
- make a v2.0.0-beta.0
- make a v2.0.0
- see the CHANGELOG for the commits between v1.0.0 and v2.0.0, not v1.0.0 -> v2.0.0-beta.0 + v2.0.0-beta.0 -> v2.0.0

### DIFF
--- a/command.js
+++ b/command.js
@@ -103,6 +103,10 @@ const yargs = require('yargs')
     type: 'string',
     describe: 'Name of the package from which the tags will be extracted'
   })
+  .option('skip-unstable', {
+    type: 'boolean',
+    describe: 'If true, unstable tags will be skipped for determining the most recent tag, e.g., x.x.x-alpha.1, x.x.x-rc.2'
+  })
   .check((argv) => {
     if (typeof argv.scripts !== 'object' || Array.isArray(argv.scripts)) {
       throw Error('scripts must be an object')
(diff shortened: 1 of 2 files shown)


ferm10n:
Just checked it out locally, these changes don't seem to be doing what I expected them to do. The bump recommendation seems to respect the flag, but not the commits included in the changelog

ferm10n:
It seems like I might also need to add a `skipUnstable` option [here](https://github.com/conventional-changelog/standard-version/blob/e4dd27c0683b97a66fe947b8ce6013e66b3fbaad/lib/lifecycles/changelog.js#L38) as well, but when I try that, the changelog output still emits a section for the alpha.

I'm not sure how to approach this yet, I'm still trying to get an idea of how conventional-changelog-core works. Maybe there is a bug in conventional-changelog-core that causes the skipUnstable to not be respected in changelog output.

ahusemann84:
I really would like to see this feature working. I am trying to setup a similar process. To achieve a similar behavior already now I use the --skip.tag option when running standard version. This way I get can get changelogs for all stable versions and at least something for the unstable ones. Although this results in a changelog for the unstable version that collects all changes since the last stable version.

PR #767, opened 2021-06-01
Combining both release-as and prerelease now doesn't break package version

### LINKED ISSUE #542: Using both prelease and release-as parameters, the version number becomes null
There was a problem when I specified a new pre-release version number

```shell
$ standard-version --release-as 4.1.0  --prerelease beta  --dry-run
√ bumping version in package.json from 4.0.0 to null
√ outputting changes to CHANGELOG.md

---

---

√ committing package.json and CHANGELOG.md
√ tagging release v4.0.0
i Run `git push --follow-tags origin master` to publish
```

kf6kjg:
Previously, as shown in #542, using both options at the same time resulted in the bump file's version being set to null.

This was validated in the first of the new unit tests. The other unit tests simply check that the rest of the behavior is correct in this case.

With this combination the user can override the version, the purpose of `release-as`, while still starting or continuing a pre-release, the purpose of of the `prerelease`.

Fixes #542

### DIFF
--- a/lib/lifecycles/bump.js
+++ b/lib/lifecycles/bump.js
@@ -27,7 +27,9 @@ async function Bump (args, version) {
   const release = await bumpVersion(args.releaseAs, version, args)
   if (!args.firstRelease) {
     const releaseType = getReleaseType(args.prerelease, release.releaseType, version)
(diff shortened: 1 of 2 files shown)


TimothyJones:

PR #788, opened 2021-08-05
feat(updater): add Gradle support

bnas:
A simple change to bring Gradle (both Groovy and Kotlin DSL format will work) support by adding the corresponding `updater`. 

This is pretty much the same code as https://github.com/conventional-changelog/standard-version/pull/591 but without the complication of adding support for Maven pom.xml.


TimothyJones:

bnas:
Ok, reopened as https://github.com/absolute-version/commit-and-tag-version/pull/34

Issue #846, opened 2021-11-03
Request: add an option to force tag replacement


pluce:
Maybe my use-case is weird but I often need to re-apply `standard-version` changelog generation when crafting my release, for many reasons: last minute bug fixes, forgotten pull requests, etc.

My workflow is:
- using `standard-version` to bump version, generate changelog, commit the file, add tag
- hell ! I forgot <something>
- merge PR / commit any content
- "fix" the release with standard-version

The problem appears on the last step, as I'd like to use `standard-version --skip.bump` to just update the Changelog without creating a new version, but it fails on the tagging step because tag already exists. I may not use `--skip.tag` because I do want the tag to be updated to the latest commit.

That's why I'd like to have an option that allows me to move the existing version tag to the latest commit.

PR incoming !

PR #847, opened 2021-11-03
feat(tag): add an option to force tag replacement

pluce:
Add a `--tag-force` flag to CLI such that `git tag` uses `-f` option. 
Useful when you re-run `standard-version` with `--skip.bump` but without `--skip.tag`.

Issues: https://github.com/conventional-changelog/standard-version/issues/846


bcoe (approved):
This seems like a good feature to me :ok: thank you for the contribution.

chrome050:
bumb

TimothyJones:

PR #903, opened 2022-05-07
Use relative path in git ignore

### LINKED ISSUE #902: Bump file are skipped using absolute path file
Hi,

`standard-version` was never bumping `package.json` on one of my project.

After some checks it looks like the file is filtered based on the absolute path of the file.
As result `dotgitignore` can incorrectly answer that a file is ignored by `.gitignore` based on parent directory not part of the project.

**How to reproduce**

- create a js project in `/home/user/myproject/` with `package.json`
- create a `.gitignore` containing `home`
- try to release a new version
- `package.json` is not updated because the abs path `/home/user/myproject/package.json` contains `home`

**Current behavior**

My `package.json` is not bumped.

**Expected behavior**

My `package.json` have to me bumped.

**Environment**
- `standard-version@9.3.2`
- `dotgitignore@2.1.0`
- Node/npm version: 7.21.0
- OS: Debian 10

**Possible Solution**

Use relative path instead of absolute path for `configFIles`

vallsv:
Closes #902 

Hi,

Here is a patch to use the relative path when checking git ignore.

The absolute path can contain stuffs which can be matched by gitignore.
It's  not an expected behavior.

I can't tell if it's the proper way to patch this problem.

### DIFF
--- a/lib/lifecycles/bump.js
+++ b/lib/lifecycles/bump.js
@@ -145,7 +145,7 @@ function updateConfigs (args, newVersion) {
     }
     const configPath = path.resolve(process.cwd(), updater.filename)
     try {
-      if (dotgit.ignore(configPath)) return
+      if (dotgit.ignore(updater.filename)) return
       const stat = fs.lstatSync(configPath)
 
       if (!stat.isFile()) return


TimothyJones:

vallsv:
Now, i am kind of afraid of the javascript software stack.
Thanks for the fork and merge.

PR #942, opened 2023-06-19
fix: initial validation of custom updaters

davidgoitia:
Custom validators passed in an options object are not validated witihn the inner object.


stevemao:
@davidgoitia can you update the tests so that we know what problems you're trying to fix?

